PR #39, opened 2021-07-06
Enhancements for direct flow 

i3149:
Bunch of bug fixes.


i3149:
Better support for direct flow.

PR #47, opened 2021-07-08
metric and metadata fixes (#46)

i3149:
* Int valued enums in metrics
* IPAddress valued metadata
* Parent/Child mapping for metadata indexes

Co-authored-by: Ian Pye <pye@kentik.com>


thezackm (dismissed):
Lgtm

PR #58, opened 2021-07-15
Adding ability to white list snmp metrics by attribute. (#57)

i3149:
Whitelist snmp attributes.

```
match_attributes: <when>
  if_Description: <matches>^igb <OR> <matches>^eth
  <OR>
  if_Alias: <matches>*foo*
```

Forcing flow timestamp to MS if its not.


thezackm (approved):
LGTM! 🚀

PR #60, opened 2021-07-15
Fix tsak (#59)

i3149:
* Fixing snmp for the tsak mock tool.

*Defaulting to not send in snmp from down interfaces (monitor_admin_shut).


thezackm (approved):
LGTM!

PR #63, opened 2021-07-16
Adding CPU and MemoryUtilization (#62)

i3149:
Also enforces a min polling time of 30 sec.

### DIFF
--- a/config/profiles/kentik_snmp/cisco/cisco-all-devices.yml
+++ b/config/profiles/kentik_snmp/cisco/cisco-all-devices.yml
@@ -79,6 +79,7 @@ metrics:
         name: cpmCPUMemoryFree
       - OID: 1.3.6.1.4.1.9.9.109.1.1.1.1.7
         name: cpmCPUTotal1minRev
+        tag: CPU
     metric_tags:
       - index: 1
         tag: cpu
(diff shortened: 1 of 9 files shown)


config/profiles/kentik_snmp/cisco/cisco-all-devices.yml
```diff
@@ -79,6 +79,7 @@ metrics:
         name: cpmCPUMemoryFree
       - OID: 1.3.6.1.4.1.9.9.109.1.1.1.1.7
         name: cpmCPUTotal1minRev
+        tag: CPU
```
thezackm, line 82: Will this override the CPU and MemoryUtilization in the Cisco Nexus profile?

thezackm (commented):
Looks good, but want to validate we're not breaking the Nexus profile

config/profiles/kentik_snmp/cisco/cisco-all-devices.yml
i3149: Great call out! What do you think about https://github.com/kentik/ktranslate/pull/63/files#diff-691ad03eaf661c8effe86119ae7cbe127095b1a18f0603c9a7d7bde018030822R522

This checks for duplicate tags and keeps the longest oid. Testing Nexus, this successfully keeps the nexus specific oid while dropping the general one.

thezackm: elegant and simple; LGTM! 🚀

thezackm (approved):

Issue #20, opened 2021-06-29
SNMP collection: ifOperStatus and ifAdminStatus wrong values


dpajin-kentik:
below is the example of the values received when running ktranslate as SNMP collector:
The values of `ifOperStatus` and `ifAdminStatus` are always 0, although the value should be 1 or 2 or possibly some other value.
I also notice that other OID values are not the exact as what I get from snmpwalk, so I concluded that since you have uptime, you are treating all OIDs as counters and supply only the difference (increase) from the previously collected value. 

I guess that the case with `ifOperStatus` and `ifAdminStatus`, since those values are not changing, they are always 0. 
Obviously your definition of IF-MIB in snmp profiles YAML files is missing the value type. Not all the OID are counters, some values are gauge, some values are timeticks and so on... this should be also configurable in YAML files.

```
        "custom_bigint": {
            "ifInErrors": 0,
            "ifInDiscards": 0,
            "ifHCOutMulticastPkts": 0,
            "Uptime": 6006,
            "ifHCInMulticastPkts": 0,
            "ifAdminStatus": 0,
            "ifOutErrors": 0,
            "ifHCInUcastPkts": 324,
            "ifOperStatus": 0,
            "ifHCInBroadcastPkts": 0,
            "ifHCOutOctets": 89290,
            "ifOutDiscards": 0,
            "ifHCOutUcastPkts": 324,
            "ifHCOutBroadcastPkts": 0,
            "ifHCInOctets": 69833
        },
        "eventType": "KSnmpInterfaceMetric",
        "provider": "kentik-router"
    },
```

In IP-MIB yaml I see there is some kind of mapping between values and strings, but this is only in IP-MIB and under the section metric tags. I have tried to do similar for `symbols` in `if-mib.yml` (see below), but so I guess it is not implementable on values. 
It might be good to have such an option for values, there are other OIDs which have mapping between integer values and the meaning, for example `bgpPeerState`.
```
  - OID: 1.3.6.1.2.1.2.2.1.7
    name: ifAdminStatus
    mapping:
      1: up
      2: down
  - OID: 1.3.6.1.2.1.2.2.1.8
    name: ifOperStatus
    mapping:
      1: up
      2: down
```

i3149:
Json is a bad format for snmp. Try running with `-format=new_relic_metric`

./bin/ktranslate -snmp snmp.yaml -format new_relic_metric -log_level info

You should get results like this, with the correct ifOperStatus and ifAdminStatus values.
```
[...]

Issue #15, opened 2021-06-28
feature - using rollup name as a measurement name for influxdb sink


dpajin-kentik:
For InfluxDB sink, all rollups data will be saved in the same measurement named "kflow".
This makes it less usable in case of usage of multiple rollups.

For example with two rollups: 
```
-rollups sum,src_addr_grouping,in_bytes,src_addr,protocol -rollups sum,dst_addr_grouping,in_bytes,dst_addr -rollup_key_join "|" 
 

> select * from "kflow" order by time desc limit 1000 
name: kflow 
time                           dst_addr        in_bytes src_addr:protocol 
----                           --------        -------- ----------------- 
2021-06-28T08:18:16.359403447Z 95.179.240.242  100       
2021-06-28T08:18:16.359403447Z 91.231.236.41   84        
2021-06-28T08:18:16.359403447Z 91.220.190.245  1002      
2021-06-28T08:18:16.359403447Z 80.81.195.249   607       
2021-06-28T08:18:16.359403447Z 8.211.41.253    119       
2021-06-28T08:18:16.359403447Z 79.99.199.251   14276     
2021-06-28T08:18:16.359403447Z 69.166.230.253  75        
2021-06-28T08:18:16.359403447Z 45.154.168.2    40        
2021-06-28T08:18:16.359403447Z 45.154.168.1    40        
2021-06-28T08:18:16.359403447Z 37.10.41.2      104       
2021-06-28T08:18:16.359403447Z 37.10.41.1      208       
2021-06-28T08:18:16.359403447Z 37.10.40.2      260       
2021-06-28T08:18:16.359403447Z 37.10.40.1      52        
2021-06-28T08:18:16.359403447Z 37.10.38.2      676       
2021-06-28T08:18:16.359403447Z 37.10.38.1      208       
2021-06-28T08:18:16.357938693Z                 128      20.151.6.228|TCP 
2021-06-28T08:18:16.357938693Z                 112      40.69.138.19|UDP 
2021-06-28T08:18:16.357938693Z                 6305164  89.21.210.84|UDP 
2021-06-28T08:18:16.357938693Z                 76       95.179.225.52|TCP 
2021-06-28T08:18:16.357938693Z                 29600    98.102.175.154|TCP 
2021-06-28T08:18:16.357938693Z                 256      40.123.207.90|UDP 
2021-06-28T08:18:16.357938693Z                 40       197.44.141.190|TCP 
[...]

PR #65, opened 2021-07-19
Supporting user tags on flow and --snmp-flow-only flag (#64)

i3149:
* Supporting user tags on flow and --snmp-flow-only flag

* #15, adding count to influx metric

* Fixing enum case for interface metrics (#20)


thezackm (approved):
LGTM! 🚀

PR #67, opened 2021-07-21
Custom Application Mapping (#66)

i3149:
* More sane way of handling asn name mapping

* Removing asn name file in favor of one in db

* finishing custom port mapping

Adds a new flag `-application_map` This overrides the static list of ports. For example:

```
applications:
  - ports: [9092,9093]
    name: kafka
  - ports: [5222, 443]
    name: my cool service
```


thezackm (approved):
LGTM! 🚀

PR #71, opened 2021-07-23
Run Docker container as non-root user

tianon:
This adds better security by default but is still pretty trivial to adjust at runtime.

(Shouldn't be any real complications to this thanks to no persistent state on-disk 💪)


i3149:
Closing in favor of #72

Issue #37, opened 2021-07-05
SNMP traps OIDs are not recognized


dpajin-kentik:
I was testing SNMP traps receiver with the linkDown and linkUp Traps and OIDs inside the traps are not recognized, although they are from IF-MIB.

Generating OIDs locally with the following commands:
```
sudo snmptrap -v2c -c public 192.168.0.130 '' IF-MIB::linkDown ifIndex i 1 ifAdminStatus i 2 ifOperStatus i 2
sudo snmptrap -v2c -c public 192.168.0.130 '' IF-MIB::linkUp ifIndex i 1 ifAdminStatus i 1 ifOperStatus i 1 
```

Results in logs:
```
2021-07-05T10:22:21.229 ktranslate [Info] KTranslate got trapdata from 127.0.0.1
[...]

PR #73, opened 2021-07-23
running ktranslate as non root;

i3149:
* Takes over from (#72)
* fixing symlink in dockerfile for /etc/mib.db -- should fix https://github.com/kentik/ktranslate/issues/37 
* fixing interface metadata for custom oids from a profile.


thezackm (approved):
lgtm!

PR #88, opened 2021-08-03
feat: adding broacade fc switch profile

thezackm:
adding new snmp profile for brocade switch

@i3149 - can you please add a new value for fibre-channel on the provider evaluation?

`provider: kentik-fibre-channel`


i3149 (approved):
LGTM

PR #90, opened 2021-08-04
Fixing brocade fc profile

i3149:
Small bug where the missing name triggers an error:

```
2021-08-03T17:01:18.129 ktranslate(557004) [Warn] KTranslate Possibly corrupted metadata table? brocade-fc-switch.yml {{1.3.6.1.2.1.75.1.2.1.1.1  map[]  } port_index  0}
```


thezackm (approved):
thanks!

PR #102, opened 2021-08-09
Bugfix: Remove a duplicated sysobjectid

ciranor:
Bugfix: Remove the .1.3.6.1.4.1.9.1.1348 sysobjectid from cisco-voice.yml, as it is also declared in the newer cisco-call-manager.yml.


i3149 (approved):
LGTM!

PR #124, opened 2021-09-02
Dynamic instrumentation.name values

i3149:
Setting `instrumentation.name` based on what profile is found for SNMP.

Also changing default snmp trap to 1620 to avoid need for root on Linux and updated windows build file.


thezackm (approved):
LGTM! 🚀

PR #125, opened 2021-09-03
Allowing override of instrumentationName on a per device basis

i3149:
set a custom `instrumentation.name` value by setting `instrumentationName` in the device snmp config.


thezackm (approved):
looks great; thanks for adding this!

PR #127, opened 2021-09-09
Fix for `match_attributes` not working

i3149:
Fixing match_attributes. These should now work fine, with and without wrapping in ". 

Also updating the error message for a broken v3 config.


thezackm (approved):
LGTM!

PR #128, opened 2021-09-10
Re-fixing bug with matching attributes, also adding some tests this time

i3149:
Covers a lot more cases with tests. Also 1 random update to snmpwalk function which supports better logging.


thezackm (approved):
Lgtm!

Issue #81, opened 2021-07-30
new_relic output format mixing attributes from multiple MIBs


dpajin-kentik:
I am polling data from multiple SNMP MIBs, concrete HOST-RESOURCES-MIB and UCD-SNMP-MIB.
The new relic output in attributes section contains some irrelevant data which comes from other MIBs, when they have the same index value.

For example: 
```
            {
                "name": "kentik.snmp.hrStorageSize",
                "type": "gauge",
                "value": 16047576,
                "attributes": {
                    "src_as_name": "Private IP",
                    "eventType": "KSnmpDeviceMetric",
                    "instrumentation.name": "snmp",
                    "key1": "aaaa",
                    "src_addr": "172.17.0.1",
                    "key2": "bbbb",
                    "laLoad": "1.16",
                    "provider": "kentik-router",
                    "src_geo": "Private IP",
                    "objectIdentifier": "1.3.6.1.2.1.25.2.3.1.5",
                    "Index": ".1",
                    "laConfig": "12.00"
                }
            },

            {
                "name": "kentik.snmp.ipSystemStatsOutFragReqds",
                "type": "gauge",
                "value": 0,
                "attributes": {
                    "laLoad": "1.16",
                    "src_as_name": "Private IP",
                    "eventType": "KSnmpDeviceMetric",
                    "key2": "bbbb",
                    "Index": ".1",
                    "instrumentation.name": "snmp",
                    "src_geo": "Private IP",
                    "key1": "aaaa",
                    "src_addr": "172.17.0.1",
                    "laConfig": "12.00",
                    "objectIdentifier": "1.3.6.1.2.1.4.31.1.1.26",
                    "provider": "kentik-router"
                }
            },
```
So `hrStorageSize` is from HOST-RESOURCES-MIB and does not have anything with the values `laLoad` and `laConfig` which are values from UCD-SNMP-MIB, but they are listed in the attributes section. Same happens with the `ipSystemStatsOutFragReqds` from the IP-MIB.

What I am missing from here is the "tag" values for example `hrStorageDescr` and `hrStorageType`, which I see that are collected initially in logs as metadata, but they are not included nowhere in the output.

```
[...]

PR #132, opened 2021-09-15
Fixing Case where attributes from multiple tables go together.

i3149:
Issue #81 

Cleans up things as long as the attribute name keeps the table name as a prefix.


thezackm:
@i3149; how would this work in the order of operations from the `tag` values? 

for instance; in the [cisco-catalyst](https://github.com/kentik/snmp-profiles/blob/a35b444d62aa977f6710a728fed88a7834e24948/profiles/kentik_snmp/cisco/cisco-catalyst.yml#L401-L407) profile we have this snippet:

```yaml
    metric_tags:
      - MIB: IF-MIB
        column:
          OID: 1.3.6.1.2.1.31.1.1.1.1
          name: ifName
        table: ifXTable
        tag: interface_name
```

will this change to match on table name by prefix happen against `name` or `tag` in this situation?

thezackm (dismissed):
🚀

nickstinemates (approved):

thezackm (approved):
lgtm

PR #133, opened 2021-09-17
snmp health metric

i3149:
Adds a seperate metric which is just for snmp health. Looks like:

```
    "metrics": [
      {
        "name": "kentik.snmp.PollingHealth",
        "type": "gauge",
        "value": 1,
        "attributes": {
          "provider": "kentik-router",
          "PollingHealth": "GOOD",
          "device_name": "bart",
          "objectIdentifier": "computed",
          "mib-name": "computed",
          "instrumentation.name": "base",
          "eventType": "KSnmpDeviceMetric"
        }
      }
    ],
```

Also fixes 1 case where provider defaults to router and not default.


thezackm (approved):
🚀  LGTM

PR #134, opened 2021-09-17
Use match_attributes to filter on device level attributes also

i3149:
This PR extends the `match_attributes` section to allow filtering on any attribute, not just interface level ones.


thezackm (approved):
LGTM

i3149:
Thanks! This is tricky code with getting `SetAttrDrop()` right. But the tests are pretty extensive so lets ship it.

PR #137, opened 2021-09-21
Build Windows Version on Main Push

i3149:
This creates a `ktranslate.msi` file as a build artifact on each main merge or snmp profile update. 

Note: Windows support is still experimental and not supported.


thezackm (approved):
🚀 🪟

PR #138, opened 2021-09-22
Building arm64 docker image also

i3149:
This builds the `kentik/ktranslate:v2arm64` docker image. It runs on arm severs, verified on AWS. 

What do you think? Should the docker tag / image be something else?

### DIFF
--- a/.github/workflows/publish-arm64.yml
+++ b/.github/workflows/publish-arm64.yml
@@ -57,6 +57,11 @@ jobs:
           repository: kentik/snmp-profiles
           path: ./config/profiles
 
+      - name: Run build
+        run: |
+          echo ${{ env.KENTIK_KTRANSLATE_VERSION }}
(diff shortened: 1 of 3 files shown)


thezackm:
@i3149 - approved; but this does bring back to forefront the discussion on moving the tags into something more standard like semantic versioning. 

Here is an example of how we tag our infrastructure agent:
Image 2021-09-22 at 8 22 12 AM

essentially each release candidate gets bundled into a single image; the latest of these also gets the `:latest` tag.

```
# Release Candidates for AMD64, ARM64, and ARM
1.20.3-rc-arm64
1.20.3-rc-arm
1.20.3-rc-amd64

# Bundle of all RCs
1.20.3-rc

# Full release
1.20.3

# Same as the full release (but only applied to the latest one of course)
latest

thezackm (approved):

i3149:
Gonna merge this in for now but we will revist tagging and such at a later date.

PR #139, opened 2021-09-23
Pye profile bulkwalk

i3149:
Allows setting `no_use_bulkwalkall` in the profile yaml.

Also fixes the windows automatic build.

### DIFF
--- a/pkg/inputs/snmp/snmp.go
+++ b/pkg/inputs/snmp/snmp.go
@@ -156,14 +156,14 @@ func runSnmpPolling(ctx context.Context, snmpFile string, jchfChan chan []*kt.JC
 		if mibdb != nil {
 			profile = mibdb.FindProfile(device.OID)
 			if profile != nil {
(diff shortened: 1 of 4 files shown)


pkg/inputs/snmp/snmp.go
```diff
@@ -160,6 +160,11 @@ func runSnmpPolling(ctx context.Context, snmpFile string, jchfChan chan []*kt.JC
 				if *dumpMibTable {
 					profile.DumpOids(cl)
 				}
+
+				if profile.NoUseBulkWalkAll {
+					device.NoUseBulkWalkAll = true
+					log.Infof("Turning off BulkWalkAll for device via profile.")
```
thezackm, line 166: does this log entry also include the `device_name` field?

i3149: Good call. Updated to add this.

thezackm (approved):

PR #144, opened 2021-09-25
Fix windows install

i3149:
Still need to figure out how to allow upgrades and not overwrite the `snmp.yaml` file.


thezackm (approved):
🚀

Issue #126, opened 2021-09-09
[Feature Request] Kentik Discovery - Device Uniqueness 


araddas:
Within our enterprise we have devices that respond to SNMP on multiple interfaces, this becomes a challenge when using discovery due to it finding the same device multiple times. The feature I am requesting is to use the SNMP EngineID for device uniqueness for Kentik Discovery, this would eliminate duplicate devices during the discovery process.

**Note there are some caveats**
1. SNMP EngineID is only available via SNMP Version 3
2. When Cisco ASA devices are configured for HA, they will report back the exact same EngineID
 
**Example response from a device responding to SNMP on multiple interfaces**
```
$ snmpwalk -v 3 10.20.30.40 .1.3.6.1.6.3.10.2.1.1.0
SNMP-FRAMEWORK-MIB::snmpEngineID.0 = Hex-STRING: 80 00 00 09 03 00 A4 B4 39 74 87 80 

$ snmpwalk -v 3 10.20.40.10 .1.3.6.1.6.3.10.2.1.1.0
SNMP-FRAMEWORK-MIB::snmpEngineID.0 = Hex-STRING: 80 00 00 09 03 00 A4 B4 39 74 87 80
```

i3149:
Makes sense. We'll add this to our Q4 roadmap.

i3149:
@araddas can you pull the latest docker image and see if this is working for you? Expected behavior is that on discovery, only one device with a unique engine ID will be present in the created yaml file.

PR #145, opened 2021-09-27
Identify devices with engineID if this is present. 

i3149:
Looking at #126

Use the EngineID to identify a device if this is returned. If two devices share a common engine id, pick the first one.

### DIFF
--- a/pkg/inputs/snmp/disco.go
+++ b/pkg/inputs/snmp/disco.go
@@ -153,7 +153,7 @@ func doubleCheckHost(result scan.Result, timeout time.Duration, ctl chan bool, m
 		}
 		md, err = metadata.GetDeviceMetadata(log, serv, nil)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/inputs/snmp/disco.go
```diff
@@ -153,7 +153,7 @@ func doubleCheckHost(result scan.Result, timeout time.Duration, ctl chan bool, m
 		}
 		md, err = metadata.GetDeviceMetadata(log, serv, nil)
 		if err != nil {
-			log.Debugf("Cannot get device metadata on %s: %v", result.Host.String(), err)
+			log.Debugf("Cannot get device metadata on %s: %v. Check for correct snmp credentials.", result.Host.String(), err)
```
thezackm, line 156: should this be moved to `Warnf` also?

thezackm (approved):
🚀 LGTM!

Issue #157, opened 2021-10-17
Enable flag for sending data to New Relic's FedRAMP authorized endpoints


nr-jbiggley:
As an engineer of a Federal institution, I would need to comply with the mandate to use FedRAMP Authorized vendor solutions. For New Relic, this would require sending data to our FedRAMP endpoints.

https://docs.newrelic.com/docs/security/security-privacy/compliance/fedramp-compliant-endpoints/#data-ingest-apis

Just as there is a `-region=US|EU` flag, having a `-fedramp=true` to force all data to those compliant endpoints would be the desired outcome.

PR #161, opened 2021-10-20
Add a GOV region for FedRAMP

i3149:
Addresses #157, easy way to send into fedramp collections. 

Usage is `-nr_region GOV` 

Results in:

```
Exporting to New Relic at main: https://gov-insights-collector.newrelic.com/v1/accounts/d/events, events: https://gov-insights-collector.newrelic.com/v1/accounts/d/events, metrics: https://gov-metric-api.newrelic.com/metric/v1, logs https://gov-log-api.newrelic.com/log/v1
```


thezackm (approved):
🚀

Issue #162, opened 2021-10-21
[Bug] match_attributes filter is not working 


araddas:
match_attributes is not working for the following filter, per https://play.golang.org/p/p-4mLKnChmS this filter should be excluding these metrics. AFAIK this pull request should have allowed this to work https://github.com/kentik/ktranslate/pull/134

```
match_attributes:
   entity_name: Sensor
```

<img width="1468" alt="Screen Shot 2021-10-21 at 1 45 21 PM" src="https://user-images.githubusercontent.com/24640633/138355452-ec2e74a9-2056-4220-a5a8-696bd3c27a6f.png">

i3149:
@araddas can you pull the latest image and test now? Thanks.

araddas:
> @araddas can you pull the latest image and test now? Thanks.

This is dropping all entity_name key/values now, the inverse of what I would have expected

**Confirming Image ID**
```
# docker images
REPOSITORY          TAG       IMAGE ID       CREATED       SIZE
kentik/ktranslate   v2        767cbcd13bbb   2 hours ago   142MB
```


**Current Match Attribute Config**
```
match_attributes:
    if_Alias: ^\*M
    entity_name: Sensor|[T-t]emp
```

**Missing entity_name attribute**
![missing_attrib](https://user-images.githubusercontent.com/24640633/139756824-5a3711e5-4d93-487a-a7df-2afb15bce6d2.png)

i3149:
`entity_name` is only valid for the `cswSwitchInfoTable` table: 
https://github.com/kentik/snmp-profiles/blob/main/profiles/kentik_snmp/cisco/cisco-all-devices.yml#L293

These was a separate bug which incorrectly applying this attribute to the entSensorValueTable. So it not being present is expected now. 

Are you enabling the `CISCO-STACKWISE-MIB`? If so, do you have the metric `kentik.snmp.cswSwitchState`? This should have `entity_name` as an attribute.

araddas:
@i3149 the stackwise mib is discovered and it worked prior to this image. I don't believe removing the entity_name is the correct thing to do, the entPhysicalDescr provides the human readable name for all of the physical devices attached to the switch (PSU, Temp Sensors, TX/RX Sensor, Etc). By removing the mapping of entity_name to the entSensorValue, you've removed the context. An example of this can be found below. You have to derive the entity name from the ENTITY-MIB and then use the index id to map that to the CISCO-ENTITY-SENSOR-MIB. 
```
[...]

PR #172, opened 2021-10-29
Looking at #162, fixing how match attributes works

i3149:
Couple major bugs fixed here:

1) User Tags were no longer being applied at all if there was a table because they didn't have a table associated. 

2) Table level attributes got applied after the match process. This resulted in random match behavior sometimes.


thezackm (approved):
🚀

PR #173, opened 2021-11-01
Build using Docker

ehazlett:
This updates the build process to build using containers.  This removes the potential for host level conflicts and makes it repeatable for local development.


.github/workflows/publish-prod.yml
```diff
@@ -64,20 +54,17 @@ jobs:
...
-
-      - name: Publish docker image
-        run: |
-          docker tag ktranslate:v2 kentik/ktranslate:v2
-          docker push kentik/ktranslate:v2
-          docker tag ktranslate:v2 kentik/ktranslate:${{ env.KENTIK_KTRANSLATE_VERSION }}
-          docker push kentik/ktranslate:${{ env.KENTIK_KTRANSLATE_VERSION }}
-          docker tag ktranslate:v2 kentik/ktranslate:latest
-          docker push kentik/ktranslate:latest
+      - name: Build and Publish Docker
```
ehazlett, line 57: @i3149 If this is OK, we will need to add a `build-arg` to use the `MAXMIND_LICENSE_KEY` env var in the Dockerfile.

i3149: Makes sense to me.

Dockerfile
```diff
@@ -1,26 +1,57 @@
...
+WORKDIR /src
+RUN make
+
+# maxmind dbs
+FROM alpine:latest as maxmind
+ARG MAXMIND_LICENSE_KEY
+RUN apk add -U curl tar
+ENV GEOLITE2_COUNTRY_FILE=GeoLite2-Country.mmdb
+ENV GEOLITE2_ASN_FILE=GeoLite2-ASN.mmdb
+RUN if [ -z "${MAXMIND_LICENSE_KEY}" ]; then echo "MAXMIND_LICENSE_KEY" not set; exit 1; fi
```
i3149, line 14: Makes sense to pull it down here.

Dockerfile
```diff
@@ -1,26 +1,57 @@
...
 	useradd --home-dir /etc/ktranslate --gid ktranslate --no-create-home --uid 1000 ktranslate
 
 COPY --chown=ktranslate:ktranslate config/ /etc/ktranslate/
+COPY --chown=ktranslate:ktranslate lib/ /etc/ktranslate/
+
+# maxmind db
+COPY --from=maxmind /GeoLite2-Country.mmdb /etc/ktranslate/
+COPY --from=maxmind /GeoLite2-ASN.mmdb /etc/ktranslate/
+# snmp
+COPY --from=snmp /snmp/profiles /etc/ktranslate/profiles/profiles
```
i3149, line 45: To clean this up a bit, possibly we should make the path just `/etc/ktranslate/profiles` without the double profiles? This was a bug I'd been meaning to get around to.

DockerfileArm64
```diff
@@ -1,29 +0,0 @@
...
-	useradd --home-dir /etc/ktranslate --gid ktranslate --no-create-home --uid 1000 ktranslate
-
-COPY --chown=ktranslate:ktranslate config/ /etc/ktranslate/
-
-# add backwards compatibility symlinks for folks using an snmp.yml from the older image (and "ls" to verify the symlinks are correct and working)
-RUN set -eux; ln -sv ktranslate/profiles ktranslate/mibs.db /etc/; ls -l /etc/profiles/ /etc/mibs.db/
-RUN set -eux; ln -sv /etc/ktranslate/mibs.db /etc/mib.db; ls -l /etc/mib.db/
-
-# Fix libpcap to work with how we built things.
-RUN set -eux; ln -sv /usr/lib/aarch64-linux-gnu/libpcap.so.1.9.1 /usr/lib/aarch64-linux-gnu/libpcap.so.1; ls -l /usr/lib/aarch64-linux-gnu/libpcap.so.1
```
i3149, line 22: Would we still need this line?

Dockerfile
```diff
@@ -1,26 +1,57 @@
...
+RUN curl -o /tmp/asn.tar.gz "https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-ASN&license_key=${MAXMIND_LICENSE_KEY}&suffix=tar.gz" && \
+	tar zxf /tmp/asn.tar.gz --strip-components 1 -C /
+
+# snmp profiles
+FROM alpine:latest as snmp
+RUN apk add -U git
+RUN git clone https://github.com/kentik/snmp-profiles /snmp
 
+# main image
+FROM ubuntu:20.04
```
i3149, line 26: Unrelated, but I've been thinking about moving to alpine from ubuntu to reduce the vulnerability scan count. Thoughts?

i3149 (approved):
LGTM -- thanks for taking this on!

DockerfileArm64
ehazlett: In my testing on my m1 the image built and ran fine using the multi-arch ubuntu.

Dockerfile
ehazlett: +1 I noticed that.  I'll remove the extra dir :)

Dockerfile
ehazlett: I'm always happy to move to Alpine :). I'll switch and see if there are any issues.

DockerfileArm64
i3149: awsome.

i3149 (approved):
:rocket:

PR #176, opened 2021-11-02
Passing KENTIK_KTRANSLATE_VERSION to build

i3149:
Also building the `:v2` tag which is what we have people pinned to in a lot of docs.


ehazlett (approved):
LGTM

ehazlett (approved):
lgtm

PR #178, opened 2021-11-03
Send 0 value metrics sometimes

i3149:
This sends metrics which are 0 valued to NR -- when they are tagged with `force=true`. 

Also fixes a bug where a format of new_relic sends metrics to the events endpoint, not the metrics one. 

Also allows forcing a particular snmp profile to be used from the yaml. 

Also lets a snmp profile use SysDesc to modify which profile is set.


thezackm (approved):
🚀

Issue #189, opened 2021-11-15
Feature: Hex to Octet conversions


Mesverrum:
Some OID's don't store IP as a string, they store it as a hex value that we are going to want to translate.
I need to take this:
.1.3.6.1.4.1.9.9.42.1.2.2.1.2.1 = Hex-String: 0A00640A
and display it as a string 10.0.100.10

From what i figured out it would be something along the lines of splitting the hex into 4 blocks then convert them each to decimal, then join them with . as delimiters

PR #191, opened 2021-11-15
Allow a hextoip conversion

i3149:
This addresses #189, converting IPs encoded as Hex into a string.


Mesverrum (approved):
Took me a bit to figure out the 4 character variant but I got there.  Looks good

PR #199, opened 2021-11-20
Allowing traps to drop unknown oids

i3149:
This is needed to support https://github.com/kentik/snmp-profiles/pull/75

Also cleaner handling of how trap yml files are processed.


thezackm (approved):
🚀

PR #201, opened 2021-11-22
Sink: GCP Pub/Sub

ehazlett:
This adds a new sink for sending to the GCP Pub/Sub service.

### DIFF
--- a/pkg/sinks/gcppubsub/gcppubsub.go
+++ b/pkg/sinks/gcppubsub/gcppubsub.go
@@ -1,0 +1,96 @@
+package gcppubsub
+
+import (
+	"context"
+	"flag"
+	"fmt"
+
+	"cloud.google.com/go/pubsub"
+	go_metrics "github.com/kentik/go-metrics"
+	"github.com/kentik/ktranslate/pkg/eggs/logger"
+	"github.com/kentik/ktranslate/pkg/formats"
+	"github.com/kentik/ktranslate/pkg/kt"
+)
+
+type GCPPubSub struct {
+	logger.ContextL
+	client    *pubsub.Client
+	registry  go_metrics.Registry
+	metrics   *GCPPubSubMetric
+	projectID string
+	topicID   string
+	topic     *pubsub.Topic
+}
+
+type GCPPubSubMetric struct {
+	DeliveryErr go_metrics.Meter
+	DeliveryWin go_metrics.Meter
+}
+
+var (
+	GCPPubSubProjectID = flag.String("gcp_pubsub_project_id", "", "GCP PubSub Project ID to use")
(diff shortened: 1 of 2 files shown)


pkg/sinks/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,96 @@
...
+
+	return nil
+}
+
+func (s *GCPPubSub) Send(ctx context.Context, payload *kt.Output) {
+	go s.send(ctx, payload.Body)
+}
+
+func (s *GCPPubSub) send(ctx context.Context, payload []byte) {
+	res := s.topic.Publish(ctx, &pubsub.Message{Data: payload})
```
i3149, line 72: I thought that gsub has keys in addition to data but I guess not.

pkg/sinks/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,96 @@
...
+	s.client = client
+	s.topic = s.client.Topic(s.topicID)
+
+	s.Infof("connected to GCP PubSub for project %s on topic %s", s.projectID, s.topicID)
+
+	return nil
+}
+
+func (s *GCPPubSub) Send(ctx context.Context, payload *kt.Output) {
+	go s.send(ctx, payload.Body)
```
i3149, line 68: I use this pattern everywhere but I worry that in an extreme case there will be too many goroutines running. I don't think worth worrying about it now.

pkg/sinks/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (s *GCPPubSub) send(ctx context.Context, payload []byte) {
+	res := s.topic.Publish(ctx, &pubsub.Message{Data: payload})
+	if _, err := res.Get(ctx); err != nil {
```
i3149, line 73: Could pass res to another channel and process one at a time but probably not worth it.

i3149 (approved):
LGMT. There's also a gcloud sink. Do you think its worth keeping this? No one uses it as far as I know.

pkg/sinks/gcppubsub/gcppubsub.go
ehazlett: I thought about this too but I think we can revisit later.  Maybe we can add some profiling as well to get some more data.

pkg/sinks/gcppubsub/gcppubsub.go
ehazlett: I think in this case it's just to confirm it succeeded.  If we wanted to process perhaps we could add a separate channel.

i3149: Makes sense. I think Publish is non blocking but Get will block until the request works/fails. Fine as is.

PR #207, opened 2021-11-30
Add elasticsearch formatter to support bulk api

ehazlett:
This adds an Elasticsearch formatter to support the bulk API.


ehazlett:
@i3149 I included the "instrumentation" and "provider" fields from the `json` formatter.  I'm not sure if we need them but I included them anyway.

i3149:
> @i3149 I included the "instrumentation" and "provider" fields from the `json` formatter. I'm not sure if we need them but I included them anyway.

Might as well keep these. Not needed but can provide some context to the data.

pkg/formats/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,141 @@
...
+		ContextL: logger.NewContextLFromUnderlying(logger.SContext{S: "elasticsearchFormat"}, log),
+	}
+
+	return jf, nil
+}
+
+func (f *ElasticsearchFormat) To(msgs []*kt.JCHF, serBuf []byte) (*kt.Output, error) {
+	msgsNew := make([]map[string]interface{}, 0, len(msgs))
+	for _, msg := range msgs {
+		if msg.EventType == kt.KENTIK_EVENT_SNMP {
```
i3149, line 39: Probobly can take this if out, just pass all types along.

pkg/formats/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,141 @@
...
+		msgsNew = append(msgsNew, mm)
+	}
+
+	esBulkData := ""
+	for _, m := range msgsNew {
+		data, err := serialize(m)
+		if err != nil {
+			return nil, err
+		}
+		esBulkData += data
```
i3149, line 53: I'm not entirely sure about how go handles string append, but generally faster to make esBulkData a slice of strings and then join with `""` at the end?

pkg/formats/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (f *ElasticsearchFormat) Rollup(rolls []rollup.Rollup) (*kt.Output, error) {
+	// serialize rolls
+	esBulkData := ""
+	for _, m := range rolls {
+		data, err := serialize(m)
+		if err != nil {
+			return nil, err
+		}
+		esBulkData += data
```
i3149, line 93: Same, avoid the looping string append.

pkg/formats/format.go
```diff
@@ -25,21 +26,24 @@ type Formatter interface {
...
+	FORMAT_NRM                  = "new_relic_metric"
+	FORMAT_SPLUNK               = "splunk"
 )
 
 func NewFormat(format Format, log logger.Underlying, compression kt.Compression) (Formatter, error) {
 	switch format {
 	case FORMAT_AVRO:
 		return avro.NewFormat(log, compression)
+	case FORMAT_ELASTICSEARCH:
+		return elasticsearch.NewFormat(log)
```
i3149, line 46: Do we want / need to handle compression here also? Not sure you ever want to gzip sending into ES but its an option.

i3149 (commented):
Looks good. Would we also want to extend the http sink to allow methods other than POST? Not sure if needed or not.

ehazlett:
> Would we also want to extend the http sink to allow methods other than POST? Not sure if needed or not.

We don't need it for the ES bulk API so I think we could have a follow up with that.

i3149 (dismissed):
LGTM

i3149 (approved):
Continues to look good, with compression added :rocket:

PR #217, opened 2021-12-14
Capping dns cache size and time spent looking

i3149:
I realized that the dns resolution feature could use some more polish. 

This caps the max number of IPs resolved at 10,000. This number can be changed by setting the `KentikMaxDnsCacheSize=10` env var. 

Also caps the time spent trying to resolve an IP. This is hard coded to 80 ms.


thezackm (approved):
🚀

Issue #222, opened 2021-12-21
🐛 BUG: bgpPeerState is receiving improper values


thezackm:
## Synopsis

We have observed across multiple environments and versions of ktranslate that the value of `bgpPeerState` is often returning outside of the [documented enumeration standard](https://github.com/librenms/librenms/blob/dc73154ae66c468890c742cf8313b23d5ddd48f0/mibs/BGP4-MIB#L195-L210) from the `BGP4-MIB`. *This is also reflected on the [bgp4-mib.yml](https://github.com/kentik/snmp-profiles/blob/0c265f9d281dcdc086f7841a174350e61c53bcc7/profiles/kentik_snmp/_general/bgp4-mib.yml#L19-L27) profile for ktranslate.*

## Issue

SNMP walks from various devices are returning valid data, with the following pattern: 

```
snmpwalk -v 2c -On -c $COMMUNITY $IP_ADDRESS .1.3.6.1.2.1.15.3.1.2
```

Results:
```
.1.3.6.1.2.1.15.3.1.2.10.3.254.6 = INTEGER: 6
.1.3.6.1.2.1.15.3.1.2.172.16.1.25 = INTEGER: 6
.1.3.6.1.2.1.15.3.1.2.172.16.2.9 = INTEGER: 6
.1.3.6.1.2.1.15.3.1.2.172.16.2.49 = INTEGER: 6
```

*Note the output pattern from this tabular OID is `OID.IP`, where `IP` is the address of the BGP peer, per the MIB. This is somewhat different than the standard SNMP table index.*

The results of the walk above show 4 BGP peers in `established: 6` status. However these are the results in New Relic:

![image](https://user-images.githubusercontent.com/30831476/147008028-ba97e188-fda9-4ed2-8bec-d96ac1a791fb.png)

Notice how there is a single entry that shows valid data; but then there are 3 extra entries with values prefixed to the `Index`. The running theory right now is that the ktranslate function that assigns `Index` values is basically creating one row per existing row in the table. So instead of each peer having a single row; each peer ends up with rows == total peer count.

You can see similar evidence here where we show all data for a single switch; and each is repeated 4X.

![image](https://user-images.githubusercontent.com/30831476/147008540-fb763d84-4707-45c3-a02d-c7c649ee349f.png)

## To Recreate

Run a current version of ktranslate against any device that supports the `BGP4-MIB`; this seems to be a universal problem based on the recurrence of log messages like this: 

```
[...]

PR #224, opened 2021-12-22
Fix device metrics

i3149:
Fix for #222 

Basically a better way of mapping mibs back to the oid which polled them.

In bgp4.yml, there are 2 oids defined:

```
      - OID: 1.3.6.1.2.1.15.3.1.2
        name: bgpPeerState
```

And

```
      - OID: 1.3.6.1.2.1.15.3.1.20
        name: bgpPeerHoldTimeConfigured
```

In the old way, code looked for the prefix of a oid to map back to a name. But, since `2` is contained by `20` the wrong enum was used. New way all the info is passed along in 1 section, so no hacky prefix lookup needed. 

Also a fix for when `mib_profile` is hand edited with a `!` to force a profile. Now this forcing is remembered across discovery runs.


thezackm (approved):
🚀 🎄 🎅

Issue #235, opened 2022-01-12
Ktranslate logs inflating due to GETBULK being used on SNMPv1 instance, despite not being supported


ASchneider-GitHub:
### Issue Summary
The customer currently has Ktranslate deployed, and is using it to get SNMP data from devices on a v1 network. The logs are inflating in size due to the error below being repeated every time a poll is run:
```
There was an SNMP polling error with the Counter walking OID <targetOID> after 0 retries: GETBULK not supported in SNMPv1.
```

### Reproduction Information
Customer currently has an SNMPv1 instance deployed on their network. They then deployed Ktranslate following the guide found here: https://docs.newrelic.com/docs/network-performance-monitoring/setup-performance-monitoring/snmp-performance-monitoring/#setup-snmp-monitoring

### Additional Information
Redacted log and configuration files have been attached:

- [c590b0cc02fa2fbe8f6f5469012eb069b7aff6d16972ff1ed214e62aab9618e8-json.log](https://github.com/kentik/ktranslate/files/7850634/c590b0cc02fa2fbe8f6f5469012eb069b7aff6d16972ff1ed214e62aab9618e8-json.log)
- [snmp-base.yaml.txt](https://github.com/kentik/ktranslate/files/7850635/snmp-base.yaml.txt)

i3149:
Good call out here. I'll update this logic.

i3149:
Should be good now with the latest build.

PR #237, opened 2022-01-15
Fix flow type, IP and snmpv1

i3149:
* Ensure that flow type and flow IP are sent to NR.

* #235 , default getbulk off when using snmpv1.


Mesverrum (approved):
Awesome,  I'd been meaning to bring that up while you were away.

Issue #245, opened 2022-01-25
💡 FR: Add `src_addr` field to `kentik-ping` devices


thezackm:
As a user, I need to track the `device_ip` of my `ping_only` devices that have `provider: kentik-ping`. Decoration of the `device_ip` into the `src_addr` field for resulting metrics can easily solve this challenge.

i3149:
Confirmed good.

Issue #244, opened 2022-01-25
💡 FR: Enhanced logging on failed metric polls


thezackm:
As a user, I need to easily find SNMP polling errors for individual metrics of assigned SNMP profiles against my target devices. 

A proposed solution is the enhance the logging available at the `Debug` level to include a new `WARN` -severity message similar to this:

```
[WARN] $DEVICE_NAME - OID failed to return results, Metric Name: $METRIC_NAME, Profile: $PROFILE_NAME
```

The reasoning for the `Debug` level log is to reduce noise on the metrics that we know will not return data based on profiles that span large amounts of device types.

i3149:
Confirmed working now.

PR #246, opened 2022-01-25
Src_addr, panic at the disco

i3149:
#244 , warns if device metrics in a profile do not return anything.
#245 , ensures src_addr is present for synthetic metrics.
Adds a guard on service_name for snmp disco to prevent a panic.
Tweak the state bot system to not fire if the right label is set.


thezackm (approved):

jbeveland27 (approved):
🚢

Issue #247, opened 2022-01-27
💡 FR: Better handling on large profile polling


thezackm:
As a user, I need the flexibility to handle large SNMP profiles in **ktranslate**. Right now, we are inheriting the hard-coded [MaxOids setting from gosnmp](https://github.com/gosnmp/gosnmp/blob/98d87374d0713c55acee86cf9b5e1f75bb5b7b83/gosnmp.go#L28), which is set at 60. However there are some profiles in the SNMP profiles repo that are exceeding this count. (ref: [netapp-cluster.yml](https://github.com/kentik/snmp-profiles/blob/main/profiles/kentik_snmp/netapp/netapp-cluster.yml) with `112` currently)

Possible solutions for this challenge:
 * Setup **ktranslate** to break up large polls into batches
 * Allow users to override the default `MaxOids` setting at Docker runtime

PR #248, opened 2022-01-28
Break big profiles up into smaller ones.

i3149:
There's a hard coded limit of 60 oids to poll at once -- bug #247.

Some big profiles have a lot more. This code breaks polls up into at most 60 polls, combining the results into a single struct.

I think the logic on the loop is right but eyes here would help.

### DIFF
--- a/pkg/inputs/snmp/metadata/device_metadata.go
+++ b/pkg/inputs/snmp/metadata/device_metadata.go
@@ -52,9 +52,36 @@ func GetDeviceMetadata(log logger.ContextL, server *gosnmp.GoSNMP, deviceMetadat
 		oids = getFromCustomMap(deviceMetadataMibs)
 	}
 
+	hasDataFull := false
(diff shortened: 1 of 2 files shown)


thezackm (approved):
🪨 ⭐ 

will our existing logging pickup this scenario? i.e.; if I have 3 batches and one fails, will I see that in the logs?

i3149:
> rock star
> 
> will our existing logging pickup this scenario? i.e.; if I have 3 batches and one fails, will I see that in the logs?

Right now, all batches have to work to be considered a success. I can change the logic so that we keep trying but report failures also?

jbeveland27 (approved):

thezackm:
> > rock star
> > will our existing logging pickup this scenario? i.e.; if I have 3 batches and one fails, will I see that in the logs?
> 
> Right now, all batches have to work to be considered a success. I can change the logic so that we keep trying but report failures also?

If the per-metric logging will still work I think we're ok; we can visit later if we find a use case for more verbosity IMO

PR #253, opened 2022-02-03
add timestamp field for formatters

ehazlett:
This adds the timestamp field for formatters that use the `util.DroppedAttrs` helper.


ehazlett:
Ah ok.  I'll move this to specific formatters instead of the general case.

Issue #257, opened 2022-02-10
🐛  - `ping_only` devices not inheriting `user_tags`


thezackm:
Currently (`kt-2022-02-09-1820292886`); devices where `ping_only: true` are not inheriting the values in either `devices.user_tags: []` or `global.user_tags: []`

Tags are being applied successfully when both `devices.response_time: true` and `global.response_time: true`, this seems to be isolated only to the `kentik-ping` devices.

PR #260, opened 2022-02-11
Reduce logging for missing metrics

i3149:
This one will only log once a missing metric at warn level, the rest of the time at debug. 

Also creates a new internal gauge metric to track these, `snmp_missing`.

Tackles #257, missing user tags as well.


thezackm (approved):
🎸 👨‍🎤

ehazlett (approved):

PR #263, opened 2022-02-14
Adding in packet loss percent and stddev time to ping measurements

i3149:
Adds in `kentik.ping.PacketLossPct` and `kentik.ping.StdDevRtt` as metrics for ping devices. 

Since there seems to be at least some interest here.


thezackm:
@314 - this is only for `ping_only: true` devices correct? not the `response_time: true` option?

i3149:
> @314 - this is only for `ping_only: true` devices correct? not the `response_time: true` option?

Should be for both. The same code is called in both places.

PR #267, opened 2022-02-16
Adding profile comments as a user tag

i3149:
Like we talked about, comments in the profile yaml files about what exact device is matching each sysoid are useful. These look like

```
sysobjectid:
  - 1.3.6.1.4.1.9.1.669     #  Cisco ASA 5510
  - 1.3.6.1.4.1.9.1.670     #  Cisco ASA 5520
  - 1.3.6.1.4.1.9.1.671     #  Cisco ASA 5520sc
```

This code grabs these out with a regex and ads it as a tag for each device. For example:

```
  "tags.kt.model": "Catalyst 9400 Series 10 Slot",
```

Questions:
* Do we want these tags on all metrics?
* Does the tag `kt.model` look good?

### DIFF
--- a/pkg/inputs/snmp/metadata/poll.go
+++ b/pkg/inputs/snmp/metadata/poll.go
@@ -66,8 +66,8 @@ func NewPoller(server *gosnmp.GoSNMP, gconf *kt.SnmpGlobalConfig, conf *kt.SnmpD
 		// See if there's a device comment for this profile.
 		dev := profile.GetDeviceSysComment(conf.OID)
 		if dev != "" {
-			log.Infof("Adding model tag %s", dev)
-			conf.AddUserTag("kt.model", dev)
+			log.Debugf("Adding model tag %s", dev)
+			conf.AddUserTag("kentik.model", dev)
 		}
 	}
 
--- a/pkg/inputs/snmp/metadata/poll.go
+++ b/pkg/inputs/snmp/metadata/poll.go
@@ -62,6 +62,13 @@ func NewPoller(server *gosnmp.GoSNMP, gconf *kt.SnmpGlobalConfig, conf *kt.SnmpD
 				}
 			}
 		}
+
+		// See if there's a device comment for this profile.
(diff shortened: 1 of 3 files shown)


jbeveland27:
These changes look good! It sounds like you chatted about this with @thezackm / @Mesverrum, so I'll defer the questions to them (I have a slight preference for the tag to be `tags.kentik.model` so it's more consistent with the `kentik.snmp.*` metric names, but I have no real stake in what it ends up being).

jbeveland27 (approved):

pkg/inputs/snmp/metadata/poll.go
```diff
@@ -62,6 +62,13 @@ func NewPoller(server *gosnmp.GoSNMP, gconf *kt.SnmpGlobalConfig, conf *kt.SnmpD
 				}
 			}
 		}
+
+		// See if there's a device comment for this profile.
+		dev := profile.GetDeviceSysComment(conf.OID)
+		if dev != "" {
+			log.Infof("Adding model tag %s", dev)
```
ehazlett, line 69: Will this get "noisy"?  Should we move it to debug?

thezackm: came here to say this 😄

thezackm (approved):
I agree with @jbeveland27 that `tags.kentik.model` would fit our pattern better; otherwise looks great!

related: this brings up the question: is it time to look into a syntax validation GHA for the `/snmp-profiles` repo now that we have work that is dependent on comments in the YAML files fitting a certain pattern?

pkg/inputs/snmp/metadata/poll.go
i3149: Its only done once at device start, but happy to move to debug as this probably isn't info the user cares about.

i3149:
I'll move to `tags.kentik.model` and ask for another round of approvals.

jbeveland27 (approved):

Issue #252, opened 2022-02-01
💡 FR: Better handling of String-valued OIDs


thezackm:
As a user, I need the ability to poll all OIDs at various intervals, regardless of their data type. Today, we place OIDs that return string values as `metric_tags` to decorate the actual integer/float `metrics` that are received on a device or SNMP table for a device. This pattern works well mot of the time, but is limited in situations where the vendor uses a string-valued OID to report status or status messages. Since `metric_tags` are polled at a 3-hour interval as "inventory" style data, this risks missing critical signals in my telemetry.

A possible work-around could be to have **ktranslate** detect string-values in `metrics` and automatically decorate them as attributes in the same way we handle `enum` today. This allows me to push my OID collection into the standard polling interval for these edge cases where I need my string-value results to stay in sync with my traditional metrics.

i3149:
Closing this one out until we need more.

PR #271, opened 2022-02-16
Basic conversion to handle string valued metrics: #252

i3149:
Is this too basic to handle this case? What I've done is added a conversion called `to_one`. For any metric, this will convert the metric to a gaudge with a value of 1. The string value will also be captured as an attribute. It will get polled every $TICK instead of every 3 hours. For example, consider a metric with 

```
      - symbols:
          OID: 1.0.8802.1.1.2.1.4.1.1.7
          name: lldpRemPortId
          conversion: to_one 
```

This will turn into a metric `snmp.lldpRemPortId` which will always equal one. It will also have an attribute called `lldpRemPortId` with the last polled value. 

Thoughts?

### DIFF
--- a/pkg/inputs/snmp/util/util.go
+++ b/pkg/inputs/snmp/util/util.go
@@ -26,6 +26,7 @@ const (
 	CONV_HEXTOIP   = "hextoip"
 	CONV_ENGINE_ID = "engine_id"
 	CONV_REGEXP    = "regexp"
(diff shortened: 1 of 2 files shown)


thezackm (approved):
🚀

Mesverrum (approved):
Looks easy enough

jbeveland27 (approved):
:shipit:

Issue #272, opened 2022-02-17
Bug: continue to poll device metadata as much as interface metadata.


i3149:
Right now, looking at the code: https://github.com/kentik/ktranslate/blob/main/pkg/inputs/snmp/metadata/poll.go#L175 we only get metadata once. Probobly this should be fetched more frequently. How often?

PR #275, opened 2022-02-18
Do not perminantly cache device metadata

i3149:
This addresses #272. Currently, device level metadata is polled once at start and then never again. 
This is a legacy behavior from when we only looked at SysOID and things like this, but doesn't seem optimal now. Patch just removes the cache part so that device metadata is polled every time interface metadata is polled. 

Separately, it looks like we're using https://github.com/kentik/ktranslate/blob/main/pkg/inputs/snmp/metadata/poll.go#L35, which is

```
	DEFAULT_INTERVAL = 30 * 60 * time.Second // Run every 30 min.
```

I really thought this was polled every 3 hours but looks like its actually every 30. What do you think? Keep at 30 or move around? Also we can change so that device metadata is polled every 3hrs, but interface 30. Thoughts?

### DIFF
--- a/pkg/inputs/snmp/metadata/device_metadata.go
+++ b/pkg/inputs/snmp/metadata/device_metadata.go
@@ -34,7 +34,7 @@ var (
 	}()
 )
 
-// Poll device-level metadata, which we only need once(?).  Works for (at least) Juniper, Cisco, and Arista.
+// Poll device-level metadata.
 func GetDeviceMetadata(log logger.ContextL, server *gosnmp.GoSNMP, deviceMetadataMibs map[string]*kt.Mib) (*kt.DeviceMetricsMetadata, error) {
 	md := kt.DeviceMetricsMetadata{
 		Customs:    map[string]string{},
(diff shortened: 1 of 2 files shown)


Mesverrum:
So just to clarify which attributes fall under device metadata?  System MIB + metric_tags?  Or is there a hardcoded listing inside ktrans?  I don't feel like 30m is too much, especially since the data gets decorated into the payload sent to NR regardless of how stale it.  30m feels like a reasonable lag time to describe to customers between making a change on a device and it updating in the console if all it costs is a dozen or so extra GETS per device

i3149:
metadata is everything under `metric_tags` which doesn't start with `if`. For example:

```
    metric_tags:
      - column:
          OID: 1.3.6.1.4.1.9.9.48.1.1.1.2
          name: ciscoMemoryPoolName
        tag: memory_pool_name
```

thezackm:
I would agree; if the current setting is 30m, then we shouldn't see a major change in data ingest by leaving that alone and adding the device metadata. I think there's a good argument to allow a customer to set metadata collection to its own interval in the `snmp-base.yaml`, but that's probably a separate FR.

thezackm (approved):

i3149:
Cool, sounds good to me.

jbeveland27 (approved):

PR #281, opened 2022-02-25
Alert if non utf8 chars are found in the output of the new relic format

i3149:
We're seeing reports where non utf8 chars can come in via snmp which when passed on to NR will trigger an `invalid json` error at the parse phase of NR (but after the web server returns 200 annoyingly enough).

This adds a basic check and will error out any payloads which contain non-uft8 chars before they get sent on. On the first error, the full payload will be sent to logging, but subsequently just the device name and provider.

For example:

```
[Error] KTranslate There was an error when converting to native: device=bart provider=kentik-net-snmp: Invalid utf8 char found in new relic output. ([{"metrics":[{"name":"kentik.snmp.ifHCOutUcastPkts","type":"count","value":0,"interval.ms":14420,"attributes":{"provider":"kentik-net-snmp","if_interface_name":"igb3","if_Netmask":"255.255.255.255","if_PhysAddress":"\u0000g#\u0019","if_Index":"4","instrumentation.name":"net-snmp" ...
``` 

And then:

```
2022-02-24T16:07:10.771 ktranslate(88802) [Error] KTranslate There was an error when converting to native: device=mabel provider=kentik-net-snmp: Invalid utf8 char found in new relic output..
```

Hopefully this should make troubleshooting these easier. Thoughts?

### DIFF
--- a/pkg/formats/nrm/nrm.go
+++ b/pkg/formats/nrm/nrm.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode/utf8"
 
 	"github.com/kentik/ktranslate/pkg/formats/nrm/events"
 	"github.com/kentik/ktranslate/pkg/formats/util"
@@ -45,6 +46,7 @@ type NRMFormat struct {
 	invalids     map[string]bool
 	mux          sync.RWMutex
 	demo         *Demozer
+	seenInvalid  bool
 
 	EventChan chan []byte
 }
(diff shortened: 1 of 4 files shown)


thezackm (dismissed):
😍  🚀

i3149:
Updated the code after a review of the data. Now will block metatdata which is bad but let the rest through. #281. 

Also fixed a bug where the profile would be stale because its only set at discovery time, not polling runtime.

thezackm (approved):

PR #282, opened 2022-03-01
Allow mulitple urls for http output; jchf input for http

i3149:
Testing out being able to chain multiple instances together to allow different sinks. 

WIP, may not go anywhere. Adds in a `/jchf` route for the http input to allow passing across instances.

### DIFF
--- a/pkg/inputs/http/http.go
+++ b/pkg/inputs/http/http.go
@@ -54,6 +54,7 @@ func NewHttpListener(ctx context.Context, host string, log logger.Underlying, re
 func (ks *KentikHttpListener) RegisterRoutes(r *kmux.Router) {
 	r.HandleFunc(Listen+"/telegraf/standard", ks.wrap(ks.readStandard))
 	r.HandleFunc(Listen+"/telegraf/batch", ks.wrap(ks.readBatch))
+	r.HandleFunc(Listen+"/ktranslate/jchf", ks.wrap(ks.readJCHF))
 }
 
 type basic struct {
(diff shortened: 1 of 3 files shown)


i3149:
I think this is now ready to look at. Lets one ktranslate instance pass traffic to n others in a fan out pattern. Idea is to allow different filters and formats for the same set of inputs. 

Also adds the flag `http_insecure` which turns off SSL checks but defaults to on. Old behavior was to always turn these off.

pkg/sinks/http/http.go
```diff
@@ -82,19 +85,20 @@ func NewSink(log logger.Underlying, registry go_metrics.Registry, sink string) (
...
 	s.client = &http.Client{Transport: s.tr}
 
 	if compression == kt.CompressionGzip {
 		s.headers["Content-Encoding"] = "GZIP"
 	}
 
-	s.Infof("Exporting to Http at %s", s.TargetUrl)
-
 	return nil
 }
```
ehazlett, line 104: Since we could potentially have several targets should we wrap the `sendHttp` context with a `WithTimeout` and specify a max to prevent dead or slow targets from stacking up goroutines?  That way we could also log it to know when a target is behaving badly?

pkg/sinks/http/http.go
```diff
@@ -82,19 +85,20 @@ func NewSink(log logger.Underlying, registry go_metrics.Registry, sink string) (
...
 
 func (s *HttpSink) Init(ctx context.Context, format formats.Format, compression kt.Compression, fmtr formats.Formatter) error {
-	s.TargetUrl = *TargetUrl
+	for _, url := range strings.Split(*TargetUrl, ",") {
```
ehazlett, line 88: Should we do a `url.Parse` here to ensure it's a valid target?

pkg/inputs/http/http.go
```diff
@@ -153,6 +154,34 @@ func (ks *KentikHttpListener) getJCHF(wrapper *basic, remoteIP string) *kt.JCHF
...
+// Get the JCHF content directly.
+func (ks *KentikHttpListener) readJCHF(w http.ResponseWriter, r *http.Request) {
+	wrapper := []*kt.JCHF{}
+
+	// Decode body in gzip format if the request header is set this way.
+	body := r.Body
+	if r.Header.Get("Content-Encoding") == "gzip" {
+		z, err := gzip.NewReader(r.Body)
+		if err != nil {
+			panic(http.StatusInternalServerError)
```
ehazlett, line 166: Any reason not to log as Error and return to keep running in case of a temporary fail instead of panic?

i3149: This is some hipster stuff I stole from @wg . These handlers are wrapped in `wrap()` here: https://github.com/kentik/ktranslate/blob/481a76e2ed93475b61bde46308009d7ff1755b04/pkg/inputs/http/http.go#L195

Uses the panic/recovery method to pass the code back to the caller but keep going. At scale I'm not sure how performant it is but it helps to simplify the code.

pkg/sinks/http/http.go
i3149: Good call.

pkg/sinks/http/http.go
i3149: Definitely.

i3149: I think I did the timeout right but I always get confused with this thing.

ehazlett: :+1:

ehazlett (approved):

PR #283, opened 2022-03-08
S3, Http config; looping polling for SNMP

i3149:
Allow snmp config files to be prefixed s3:// or http(s):// in addition to a local path. If a remote path, fetch the resource from the remote place. 

Also, adds a flag `--snmp_discovery_min` which will run discovery every X minutes and restart the snmp polling system if a new device is found.

### DIFF
--- a/pkg/inputs/snmp/disco.go
+++ b/pkg/inputs/snmp/disco.go
@@ -275,7 +275,7 @@ func addDevices(ctx context.Context, foundDevices map[string]*kt.SnmpDeviceConfi
 			added++
 		} else if conf.Devices[key] != nil {
 			if conf.Disco.ReplaceDevices { // But keep backwards compatible with existing devices and don't change their entries.
-				d.UpdateFrom(conf.Devices[key])
+				d.UpdateFrom(conf.Devices[key], conf)
 				conf.Devices[key] = d
 				replaced++
 			} else {
@@ -283,7 +283,7 @@ func addDevices(ctx context.Context, foundDevices map[string]*kt.SnmpDeviceConfi
 			}
 		} else {
 			if conf.Disco.ReplaceDevices { // Else, new style keys all use keyAlt.
-				d.UpdateFrom(conf.Devices[keyAlt])
+				d.UpdateFrom(conf.Devices[keyAlt], conf)
 				conf.Devices[keyAlt] = d
 				replaced++
 			} else {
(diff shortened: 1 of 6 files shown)


pkg/inputs/snmp/disco.go
```diff
@@ -248,7 +276,7 @@ func doubleCheckHost(result scan.Result, timeout time.Duration, ctl chan bool, m
 	foundDevices[result.Host.String()] = &device
 }
 
-func addDevices(foundDevices map[string]*kt.SnmpDeviceConfig, snmpFile string, conf *kt.SnmpConfig, isTest bool, log logger.ContextL) error {
+func addDevices(ctx context.Context, foundDevices map[string]*kt.SnmpDeviceConfig, snmpFile string, conf *kt.SnmpConfig, isTest bool, log logger.ContextL) (error, int, int, int) {
```
ehazlett, line 279: Should we do something like create a struct for the return?  Something like
```type snmpDevices {
      added int
      replaced int
      delta int
}
```
and then return `(*snmpDevices, error)`?  It makes it a bit more readable IMO.  I know I'll forget this order down the road ;)

pkg/inputs/snmp/snmp.go
```diff
@@ -33,6 +33,7 @@ var (
 	snmpWalkFormat = flag.String("snmp_walk_format", "", "use this format for walked values if -snmp_do_walk is set.")
 	snmpOutFile    = flag.String("snmp_out_file", "", "If set, write updated snmp file here.")
 	snmpPollNow    = flag.String("snmp_poll_now", "", "If set, run one snmp poll for the specified device and then exit.")
+	snmpDiscoDur   = flag.Int("snmp_discovery_sec", 0, "If set, run snmp discovery on this interval (in seconds).")
```
thezackm, line 36: recc updating this to `snmp_discovery_min` to allow a more user friendly scheduling experience

i3149: good call.

pkg/inputs/snmp/disco.go
i3149: hah, totally.

i3149:
OK -- thanks for the pre-comments. I think this is ready to look at generally.

ehazlett (approved):

i3149:
@araddas -- any opinions on this one?

araddas:
@i3149 im on PTO, will get back to you next week.

araddas:
@i3149 this looks good, shoot off a build and I will test it out.

jbeveland27 (approved):

i3149:
> @i3149 this looks good, shoot off a build and I will test it out.

Cool, give it a shot now when you get a chance.

Issue #284, opened 2022-03-11
Bug: syslog sometimes missing device_name/hostname


jbeveland27:
@jbiggley was looking through syslog data and noticed that `device_name` and `hostname` are empty (Josh - please correct if that's wrong). When that occurs the log is relatively meaningless since there's no way to know which device it was associated with.

There's a fallback present [here](https://github.com/kentik/ktranslate/blob/main/pkg%2Finputs%2Fsyslog%2Fsyslog.go#L188-L193) where the hostname is used if the device_name is empty, but it might be more robust to do a dns lookup if the dns flag is set.

i3149:
Should we enforce `hostname` being populated always? Right now, this may or may not be populated, but `device_name` always should be.

jbeveland27:
`device_name` is good. Thanks!

PR #285, opened 2022-03-11
Trying extra hard to have device_name always set for syslog

i3149:
I think this should address #284 

Works extra hard to make sure that `device_name` is filled in and set. Note -- works best if the `-dns=local` flag is set. Worth it to always set this for syslog listeners?

### DIFF
--- a/pkg/cat/kkc.go
+++ b/pkg/cat/kkc.go
@@ -28,6 +28,7 @@ import (
 	"github.com/kentik/ktranslate/pkg/sinks/kentik"
 	"github.com/kentik/ktranslate/pkg/util/enrich"
 	"github.com/kentik/ktranslate/pkg/util/gopatricia/patricia"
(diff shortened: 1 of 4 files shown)


jbeveland27 (approved):

jbeveland27:
> Worth it to always set this for syslog listeners?
I have an open question internally on if we want to always set `-dns=local` in our setup wizard. I'm assuming it'll be a yes, but just wanted to say we're following up on it.

i3149:
> > Worth it to always set this for syslog listeners?
> 
> I have an open question internally on if we want to always set `-dns=local` in our syslog setup wizard. I'm assuming it'll be a yes, but just wanted to say we're following up on it.

Cool. Doing dns takes a little bit of CPU and memory for the cache. Probably yes but may want a way to turn off if there's scaling issues.

Issue #288, opened 2022-03-14
Bug: response_time PacketLossPct reporting nonsense values


Mesverrum:
I ran into this in a customer account and was able to duplicate it in my test environment, with reponse_time: true globally  kentik.ping.PacketLossPct seems to always be returning  -9223372036854776.  I would expect this should be 0 in the places I tested it.

i3149:
This is an artifact of when the ping is not getting returned. Likely will be fixed by setting this env var: `KENTIK_PING_PRIV=true` 

This changes ping from UDP to ICMP based packets.

drhythm:
Hi @i3149 is this a change that the customer would make in their environment, or should we wait for a new container version to be released?

Also, I just want to confirm how this works... so the desired output = 0.. and if the returned value is more than 0, that means the node is down / not responding to ping ?

Issue #286, opened 2022-03-11
💡 FR - Add support for multiple metrics with single tag in profiles


thezackm:
As a user, I need the SNMP profiles to be flexible and allow for situations where multiple devices in a single category use different OIDs for a single metric. 

For example; Dell PowerConnect typically uses the `DNOS-SWITCHING-MIB` for the PowerConnect series of switches, but we've also received SNMP walks on a model which uses the `RADLAN-MIB`. 

The expectation is that ktranslate can identify which of these OIDs returns a value on polling, and forward that one to the nrSink. 

ex:
```yaml
metrics:
  # This OID is for CPU on the PowerConnect 3548
  - MIB: RADLAN-MIB
    symbol:
      name: rlCpuUtilDuringLastMinute
      OID: 1.3.6.1.4.1.89.1.8.0
      poll_time_sec: 60
      tag: CPU

  # Total CPU utilization
  - MIB: DNOS-SWITCHING-MIB
    symbol:
      name: agentSwitchCpuProcessPercentageUtilization
      OID: 1.3.6.1.4.1.674.10895.5000.2.6132.1.1.1.1.4.9.0
      poll_time_sec: 60
      tag: CPU
```

PR #290, opened 2022-03-15
Not sending in Ping PacketLoss if the value is NaN: #288

i3149:
Fixing #288 

UDP ping can fail. In this case, you get NaN for PacketLoss. This PR will not set a value if ping is not working. 

Q-- should there be a metric or other way to indicate that ping is broken in this case?

### DIFF
--- a/pkg/inputs/snmp/mibs/profile.go
+++ b/pkg/inputs/snmp/mibs/profile.go
@@ -26,6 +26,7 @@ type OID struct {
 	PollTime   int              `yaml:"poll_time_sec,omitempty"`
 	MatchAttr  []string         `yaml:"match_attributes,omitempty"`
 	Format     string           `yaml:"format,omitempty"`
+	AllowDup   bool             `yaml:"allow_duplicate,omitempty"`
 }
 
 type Tag struct {
@@ -385,6 +386,7 @@ func (p *Profile) GetMetrics(enabledMibs []string, counterTimeSec int) (map[stri
 				Table:        metric.Table.GetTableName(),
 				FromExtended: metric.fromExtended,
 				Format:       metric.Symbol.Format,
+				AllowDup:     metric.Symbol.AllowDup,
 			}
 			if len(mib.Enum) > 0 {
 				mib.EnumRev = make(map[int64]string)
@@ -426,6 +428,7 @@ func (p *Profile) GetMetrics(enabledMibs []string, counterTimeSec int) (map[stri
 				Table:        metric.Table.GetTableName(),
 				FromExtended: metric.fromExtended,
 				Format:       s.Format,
+				AllowDup:     s.AllowDup,
 			}
 			if len(mib.Enum) > 0 {
 				mib.EnumRev = make(map[int64]string)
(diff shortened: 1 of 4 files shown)


i3149:
Also joined in #286 

Adds in a `allow_duplicate` field to snmp yaml which lets two+ metrics share a tag. Only metrics which return a value when polling will be used.  

For example:

```
metrics:
  # This OID is for CPU on the PowerConnect 3548
  - MIB: RADLAN-MIB
    symbol:
      name: rlCpuUtilDuringLastMinute
      OID: 1.3.6.1.4.1.89.1.8.0
      poll_time_sec: 60
      tag: CPU
      allow_duplicate: true

  # Total CPU utilization
  - MIB: DNOS-SWITCHING-MIB
    symbol:
      name: agentSwitchCpuProcessPercentageUtilization
      OID: 1.3.6.1.4.1.674.10895.5000.2.6132.1.1.1.1.4.9.0
      poll_time_sec: 60
      tag: CPU
      allow_duplicate: true
```

thezackm (approved):

pkg/inputs/snmp/mibs/profile.go
```diff
@@ -777,7 +782,7 @@ func prune(mibs map[string]*kt.Mib, counterTimeSec int, minCounterTime int) {
 
 	// Lastly, delete if we're not keeping.
 	for oid, mib := range mibs {
-		if !keepNames[oid] {
+		if !keepNames[oid] && !mib.AllowDup {
```
jbeveland27, line 785: > Only metrics which return a value when polling will be used.

Question: it looks like the `allow_duplicate` field is just acting as an override so that the oid is kept as a metric to be polled. But I'm not noticing if the the metric was polled to make sure a value returned. If an OID is polled and nothing is returned, is that just omitted when the metrics get sent?

jbeveland27:
@thezackm more of a philosophical question: say two metrics had the same tag and _both_ returned values. Would that be an issue? I didn't notice anything in the code really guarding against that scenario, so just curious if that's something we need to care about.

thezackm:
> @thezackm more of a philosophical question: say two metrics had the same tag and _both_ returned values. Would that be an issue? I didn't notice anything in the code really guarding against that scenario, so just curious if that's something we need to care about.

ironically we just started a thread on that in Slack - we might need to come up with ***another*** key to mandate which OID takes priority. It should be a VERY edge case though (in this case, accounting for devices that use both 32-bit and 64-bit counters in the IF-MIB).

pkg/inputs/snmp/mibs/profile.go
i3149: In theory, what happens is first all of the oid get polled: https://github.com/kentik/ktranslate/blob/main/pkg/inputs/snmp/metrics/device_metrics.go#L93

Then, for those that did return a result, we walk over the results and do stuff with them:
https://github.com/kentik/ktranslate/blob/main/pkg/inputs/snmp/metrics/device_metrics.go#L146

What I'm hoping happens here is that this OID will get polled, but since nothing is returned, it will just get dropped in the 2nd loop and not sent back home.

i3149:
Added one more commit to fix the build I broke with a bunch of other prs put together.

thezackm (approved):

PR #295, opened 2022-03-18
Fixing a crash caused by a missing mib in metadata

i3149:
Most recent release has a bug where if a metadata oid is polled and is missing a mib for it the container will crash:

```
goroutine 75 [running]:
github.com/kentik/ktranslate/pkg/kt.(*Mib).IsPollReady(0xc002271b38)
        /src/pkg/kt/snmp.go:312 +0x18
github.com/kentik/ktranslate/pkg/inputs/snmp/metadata.(*DeviceMetadata).poll(0xc001f664c0, {0x181c300, 0xc00116a000}, 0x0)
```

This ads a nil check on the offending line, causing the system to work as normal.


thezackm (approved):

Mesverrum (approved):

jbeveland27 (approved):
:shipit:

PR #302, opened 2022-03-30
Changing ping logic to always run, even when device is offline

i3149:
There's two ways that ping can get turned on:

1) set `ping_only: true` in the device config.
2) set `response_time: true` in the global section.

If the device is down, currently ping gets turned off along with the rest of the snmp checks unless its set up in mode 1. 
This PR changes mode 2 so that it works like mode 1 and always pings the target device, even if the device is down.

### DIFF
--- a/pkg/inputs/snmp/metrics/device_metrics.go
+++ b/pkg/inputs/snmp/metrics/device_metrics.go
@@ -271,26 +271,6 @@ func (dm *DeviceMetrics) pollFromConfig(ctx context.Context, server *gosnmp.GoSN
 		dst.CustomStr["profile_message"] = kt.DefaultProfileMessage
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/inputs/snmp/metrics/device_metrics.go
```diff
@@ -271,26 +271,6 @@ func (dm *DeviceMetrics) pollFromConfig(ctx context.Context, server *gosnmp.GoSN
 		dst.CustomStr["profile_message"] = kt.DefaultProfileMessage
 	}
```
i3149, line 273: Don't need this code block any more because the measurements come in via a separate thread.

Mesverrum (approved):

thezackm (approved):

Issue #300, opened 2022-03-25
💡 FR: Add "reason" to PollingHealth metric


thezackm:
As a user, it's important for me to understand why my devices end up with `PollingHealth = 'BAD'` values. As of today, the understanding is the `BAD` threshold can be triggered by the following:
  - Polling takes longer than the configured `poll_time_sec`
  - No packets received from SNMP query
  - No metrics parsed from SNMP response

To support this use case, we need to decorate the `kentik.snmp.PollingHealth` metric with an additional attribute to explain the reason for the current polling status. That will enable queries like this in NRQL (assuming the new field is named `HealthReason`):

```
FROM Metric 
SELECT latest(timestamp), latest(PollingHealth), latest(HealthReason) 
FACET device_name, instrumentation.name AS 'Profile', provider 
WHERE kentik.snmp.PollingHealth[latest] = 2 
LIMIT MAX
```

PR #303, opened 2022-03-30
Update health with status of why bad

i3149:
#300 

This adds a reason to the polling health metric.

Right now, there are these reasons:

```
		SNMP_GOOD:                   "GOOD",
		SNMP_BAD_NO_FLOWS:           "BAD: No flows",
		SNMP_BAD_DEVICE_ERR:         "BAD: Device Error",
		SNMP_BAD_INTERFACE_ERR:      "BAD: Interface Error",
		SNMP_BAD_POLL_TIMEOUT:       "BAD: Poll Timeout",
		SNMP_BAD_INIT_METADATA:      "BAD: Init Metadata",
		SNMP_BAD_INIT_METRICS:       "BAD: Init Metrics",
		SNMP_BAD_FIRST_METRICS_POLL: "BAD: Metrics Poll",
```

If bad, these are separated into 2 attributes:

```
"PollingHealth": "BAD",
```
And
```
"PollingHealthReason": "No flows"
```

If good, `PollingHealthReason` is not set. 

Other trick is that the value can go from 

```
	SNMP_GOOD                   = 1
	SNMP_BAD_NO_FLOWS           = 3
	SNMP_BAD_DEVICE_ERR         = 4
	SNMP_BAD_INTERFACE_ERR      = 5
	SNMP_BAD_POLL_TIMEOUT       = 6
	SNMP_BAD_INIT_METADATA      = 7
	SNMP_BAD_INIT_METRICS       = 8
	SNMP_BAD_FIRST_METRICS_POLL = 9
```

Where anything > 1 is bad. Should it be collapsed down into the 1 or 2 range again?

### DIFF
--- a/pkg/kt/snmp.go
+++ b/pkg/kt/snmp.go
@@ -238,7 +238,6 @@ type SnmpDeviceMetric struct {
 
 const (
 	SNMP_GOOD                   = 1
-	SNMP_BAD                    = 2
 	SNMP_BAD_NO_FLOWS           = 3
 	SNMP_BAD_DEVICE_ERR         = 4
 	SNMP_BAD_INTERFACE_ERR      = 5
--- a/pkg/kt/snmp.go
+++ b/pkg/kt/snmp.go
@@ -237,14 +237,28 @@ type SnmpDeviceMetric struct {
 }
 
 const (
-	SNMP_GOOD = 1
-	SNMP_BAD  = 2
+	SNMP_GOOD                   = 1
+	SNMP_BAD                    = 2
+	SNMP_BAD_NO_FLOWS           = 3
+	SNMP_BAD_DEVICE_ERR         = 4
+	SNMP_BAD_INTERFACE_ERR      = 5
+	SNMP_BAD_POLL_TIMEOUT       = 6
+	SNMP_BAD_INIT_METADATA      = 7
+	SNMP_BAD_INIT_METRICS       = 8
+	SNMP_BAD_FIRST_METRICS_POLL = 9
 )
 
 var (
 	SNMP_STATUS_MAP = map[int64]string{
-		1: "GOOD",
-		2: "BAD",
+		SNMP_GOOD:                   "GOOD",
+		SNMP_BAD:                    "BAD",
+		SNMP_BAD_NO_FLOWS:           "BAD: No flows",
+		SNMP_BAD_DEVICE_ERR:         "BAD: Device Error",
+		SNMP_BAD_INTERFACE_ERR:      "BAD: Interface Error",
+		SNMP_BAD_POLL_TIMEOUT:       "BAD: Poll Timeout",
+		SNMP_BAD_INIT_METADATA:      "BAD: Init Metadata",
+		SNMP_BAD_INIT_METRICS:       "BAD: Init Metrics",
+		SNMP_BAD_FIRST_METRICS_POLL: "BAD: Metrics Poll",
 	}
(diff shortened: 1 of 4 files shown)


pkg/kt/snmp.go
```diff
@@ -237,14 +237,26 @@ type SnmpDeviceMetric struct {
...
+	SNMP_BAD_INIT_METRICS       = 8
+	SNMP_BAD_FIRST_METRICS_POLL = 9
 )
 
 var (
 	SNMP_STATUS_MAP = map[int64]string{
-		1: "GOOD",
-		2: "BAD",
+		SNMP_GOOD:                   "GOOD",
+		SNMP_BAD_NO_FLOWS:           "BAD: No flows",
```
thezackm, line 253: this language leads me to think my NetFlow flows are broken; but I think the actual measurement is against the amount of data in the response payload from SNMP? 

If that's accurate; I wonder if we should use something more like `SNMP_BAD_NO_DATA: "BAD: No data"`

i3149: Good call. Updated.

thezackm (approved):

pkg/kt/snmp.go
```diff
@@ -237,14 +237,26 @@ type SnmpDeviceMetric struct {
...
-	SNMP_GOOD = 1
-	SNMP_BAD  = 2
+	SNMP_GOOD                   = 1
+	SNMP_BAD_NO_DATA            = 3
+	SNMP_BAD_DEVICE_ERR         = 4
+	SNMP_BAD_INTERFACE_ERR      = 5
+	SNMP_BAD_POLL_TIMEOUT       = 6
+	SNMP_BAD_INIT_METADATA      = 7
+	SNMP_BAD_INIT_METRICS       = 8
+	SNMP_BAD_FIRST_METRICS_POLL = 9
```
jbeveland27, line 247: nit: is there a reason `2` was skipped? This goes 1 -> 3,4,5...9

i3149: Old pattern was just 

```
const (
	SNMP_GOOD = 1
	SNMP_BAD  = 2
)
```

Taking out SNMP_BAD and I didn't want to confuse old data. Could keep things 1,2,3... though as this is kinda weird. Thoughts?

jbeveland27: Oh hmm, the way I read this was that the actual data set in `kentik.snmp.PollingHealth` came from `SNMP_STATUS_MAP`, so the actual values were only ever `GOOD`|`BAD`. The numbers are just to represent the enum, right? Or am I reading this wrong 🥴

i3149: That's the old way. The new way (which is totally a hack) still gets the string values from SNMP_STATUS_MAP but splits the string into 2 parts, one which is `GOOD|BAD` and the other is the reason. But the int part which the guadge value is set to is either 1 or 3-9 depending on the actual reason.

jbeveland27: >But the int part which the guadge value is set to is either 1 or 3-9 depending on the actual reason.
Ah, I didn't realize the gauge is set from this. Makes sense when you put it that way 👍🏻

jbeveland27 (approved):

pkg/kt/snmp.go
i3149: I was gonna try and keep it 1 or 2 but then it seemed easy enough to alert on 1 or not 1.

PR #308, opened 2022-04-05
Fixing nulls for rfc5424 style messages

i3149:
This fixes a case where `NULLs` would come in for some syslog messages `message` field. 

Bug is that `content` was swapped for `message` field because `message` is what NR looks for. But in RFC5424, its already the message field and there is no content field. Fix is to be smarter about this and only swap if there both is NO message and there is a content.


thezackm (approved):

jbeveland27 (approved):
Nice catch. LGTM

PR #314, opened 2022-04-20
chore: update add_mibs to default true

thezackm:
requesting update to default YAML files to change `add_mibs: true` for discovery settings. we've had a few instances of users pulling their YAML directly off the container and not updating this setting before running discovery and then having confusion about missing metrics later.


Mesverrum (approved):

jbeveland27 (approved):
🚢

i3149 (approved):
LGTM

PR #318, opened 2022-04-27
Example deployment with snmp and snmp trap listener in k8s sending to nr

i3149:
Making a quick demo service showing how to send data into NR via k8s. 

Uses secrets to store account ID and API KEY.


thezackm (approved):
🧊

i3149:
thanks for the reviews!

Issue #315, opened 2022-04-25
FR: lookup new flow entities by DNS if possible


i3149:
When creating a new entity, system will use sender's IP if the name is not known. 

Add a step where reverse dns is tried before falling back to just IP.

PR #321, opened 2022-04-28
Resolve flow sending devices by dns

i3149:
If a device's IP is not in `snmp.yaml`, netflow collector doesn't know what to call it. This will use dns if possible to resolve it to a device name.

Also adds locking to the resolver class which probably should have been there the whole time. 

#315

### DIFF
--- a/pkg/inputs/flow/format.go
+++ b/pkg/inputs/flow/format.go
@@ -15,6 +15,7 @@ import (
 	"github.com/kentik/ktranslate/pkg/eggs/logger"
 	"github.com/kentik/ktranslate/pkg/kt"
 	"github.com/kentik/ktranslate/pkg/util/ic"
+	"github.com/kentik/ktranslate/pkg/util/resolv"
 
 	flowmessage "github.com/netsampler/goflow2/pb"
 )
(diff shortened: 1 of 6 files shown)


pkg/inputs/flow/format.go
```diff
@@ -106,6 +111,12 @@ func (t *KentikDriver) toJCHF(fmsg *flowmessage.FlowMessage) *kt.JCHF {
 		dev.SetUserTags(in.CustomStr)
 	} else {
 		in.DeviceName = net.IP(fmsg.SamplerAddress).String()
+		if t.resolv != nil {
+			dm := t.resolv.Resolve(t.ctx, in.DeviceName)
+			if dm != "" {
+				in.DeviceName = dm
+			}
+		}
```
thezackm, line 119: @i3149 - can we add a one-time `Info` level log message that shows the results of this check?

```
$src_addr resolved to $device_name
```

something along those lines? just so we have tracking on the attempt.

i3149: definitely.

thezackm (approved):
🪨 ⭐

PR #323, opened 2022-05-20
elasticsearch: include more geo data

ehazlett:
This adds more geolocation fields to the elasticsearch fomatter.

### DIFF
--- a/pkg/formats/elasticsearch/elasticsearch.go
+++ b/pkg/formats/elasticsearch/elasticsearch.go
@@ -192,8 +192,12 @@ var (
 	}
 
 	KeepAttrs = map[string]bool{
-		"lost": true,
-		"sent": true,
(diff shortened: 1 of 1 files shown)


i3149 (approved):
This passes these fields regardless, even if they are not set. But if they are set, the fields should still be going through, not just getting dropped right?

ehazlett:
> This passes these fields regardless, even if they are not set. But if they are set, the fields should still be going through, not just getting dropped right?

Correct.  Should I check if they are empty or leave them as passing?  I'm not sure what convention we use for other formatters and what's preferred.

ehazlett:
Ah I just saw `// Always pass along even if 0` -- should we leave it to include the fields then?

i3149:
Eah, can't hurt to have.

PR #325, opened 2022-05-24
Add more examples

ehazlett:
This adds more examples including an example Lima VM and Elasticsearch, Prometheus, InfluxDB, and an SNMP example using the Lima VM.


i3149 (approved):
:dollar: :dollar: :dollar:

PR #329, opened 2022-06-02
Tweaking how bgp version and intrumentation.name get set

i3149:
BGP Version doesn't change so makes sense to make it an attribute.

Updating synth intrumentation.name to allow special handing for different test types.


ehazlett (approved):
lgtm

PR #339, opened 2022-06-16
Influx batches should end with newline

tdanner:
This may not matter for sending influx over http which sends each batch as a separate request, but when you send it to a file some of the influx lines get jammed together without a newline separator. This fixes that.

PR #340, opened 2022-06-16
Influx escape

tdanner:
Escape special characters according to influx line format docs

This allows tag values containing space, comma, and equals to be handled.

Note: this PR also includes the changes from #339


i3149 (approved):
LGTM!

PR #341, opened 2022-06-16
Make rpm

i3149:
Super basic rpm which drops the min required files onto a server. 

Currently only rels8.

### DIFF
--- a/scripts/post-install
+++ b/scripts/post-install
@@ -1,0 +1,1 @@
+git clone https://github.com/kentik/snmp-profiles.git /etc/ktranslate/profiles
--- a/scripts/post-install
+++ b/scripts/post-install
@@ -1,1 +1,6 @@
-su -s /bin/bash ktranslate -c "git clone https://github.com/kentik/snmp-profiles.git /etc/ktranslate/profiles"
+if [ -d "/etc/ktranslate/profiles" ]
+then
+  su -s /bin/bash ktranslate -c "cd /etc/ktranslate/profiles && git pull --rebase"
+else
+  su -s /bin/bash ktranslate -c "git clone https://github.com/kentik/snmp-profiles.git /etc/ktranslate/profiles"
+fi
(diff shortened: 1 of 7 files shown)


i3149:
rpm to test is here: https://github.com/kentik/ktranslate/actions/runs/2511490983.

This then drops a ktranslate service file which can be run with: 

sudo systemctl start ktranslate

i3149:
Also will need to clone snmp profiles if needed, do with

```
sudo -u ktranslate git clone https://github.com/kentik/snmp-profiles.git /etc/ktranslate/profiles
```

i3149:
Now building rels 7,8 repos
https://github.com/kentik/ktranslate/actions/runs/2511776100

i3149:
And debian/ubuntu: https://github.com/kentik/ktranslate/actions/runs/2511886286

scripts/post-install
```diff
@@ -0,0 +1 @@
+su -s /bin/bash ktranslate -c "git clone https://github.com/kentik/snmp-profiles.git /etc/ktranslate/profiles"
```
ehazlett, line 1: Could we just include this as part of the package instead of having to have a post script to pull?

ehazlett (dismissed):
Minor nit other LGTM!

scripts/post-install
i3149: Great q -- I went around on this for a while and couldn't get it to work. Specifically, https://github.com/kentik/pkg doesn't support globs for which files to get put into a package. We could script the add of each file in the workflow, but that seemed a bit messy. 

Separately, pulling the snmp profiles out of here make versioning ktrans code a lot easier. No need for a new release every time a snmp profile gets changed. Con is that it pushes keeping up to date here onto users, but that seems reasonable. 

Typing this made me realize though that installing a new version of ktrans will error out because the git repo will already exist. Time for a more complex script which knows to git update in this case I think.

ehazlett: Got it -- makes sense :+1:

ehazlett:
Only other thing would be the package name for other distros that don't use `git-core` (i.e Alpine) but we don't have that as a target so should be fine for now.

LGTM!

PR #343, opened 2022-06-22
Release tags

i3149:
This PR sets up release packaging so that users can download rpms directly from this github: https://github.com/kentik/ktranslate/releases/tag/v1.0.0-rc6 as an example.

Also, it creates a new ktranslate@ service. This allows commands like `sudo systemctl start ktranslate@flow` where the flow part allows seperate ktranslate instances per server. Works with 

```
EnvironmentFile=-/etc/default/ktranslate-%i.env
```

Where anything after the `@` goes into the `%i`  variable, which loads a specific env var.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -94,6 +94,17 @@ jobs:
           format: ${{ matrix.pkg }}
           package: ktranslate-package.yml
 
+      - name: Upload Release Asset
+        id: upload-release-asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ github.event.release.upload_url }}
+          asset_path: ${{ steps.package.outputs.package }}
+          asset_name: ${{ steps.package.outputs.package }}--${{ matrix.distro }}-${{ matrix.release }}
+          asset_content_type: application/binary
+
       - name: Send to packagecloud
         shell: bash
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: Upload Release Asset
+        id: upload-release-asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ github.event.release.upload_url }}
+          asset_path: ${{ steps.package.outputs.package }}
+          asset_name: ${{ steps.package.outputs.package }}--${{ matrix.distro }}-${{ matrix.release }}
+          asset_content_type: application/binary
```
jbeveland27, line 116: One note here: this uploads an asset, but what creates the release? Is that done within the `kentik/pkg` action?

jbeveland27:
Generally speaking I think this is great. I'm still toying with goreleaser separately as a more all-inclusive release tool that would also auto-generate changelog docs for us as well. But I think this gets things going really quickly

jbeveland27 (approved):

.github/workflows/create-release.yml
i3149: Right now, its a manual process to create a release. I click around in the GH webite.

i3149:
Lots of pain later, and things are also pushing to package cloud. 

https://packagecloud.io/kentik/ktranslate

This lets you do things like `apt-get install ktranslate` and `yum install ktranslate`

jbeveland27 (approved):
Looks great 🍪

qryuu:
About `@flow`
Ktranslate needs to be separated for each FlowType.
`@sflow` and `@netflow5` etc.

Issue #337, opened 2022-06-15
FR -- Gracefully Handle Devices which are no longer present.


i3149:
The right approach might be to offer both options. `delete_after` could be per-device and there could be something like a `global.purge_devices: true` that would remove devices if they're not discovered. Gives more flexibility to the users to decide (edited)

PR #344, opened 2022-06-23
#337 -- allow an option to remove non polling snmp devices automatically

i3149:
Looking at #337, ways to allow non found devices to be removed. 

Adding a config `global.purge_devices_after_num` which means that after N times that a device isn't seen in snmp polling, it will be removed. Doesn't do anything by default. Also right now will not do anything in a normal discovery, only auto-removes when the `-snmp_discovery_min` flag is set. 

Q: Is this what we want? Possibly should also remove on a normal discovery?

### DIFF
--- a/pkg/inputs/snmp/disco.go
+++ b/pkg/inputs/snmp/disco.go
@@ -369,12 +369,19 @@ func addDevices(ctx context.Context, foundDevices map[string]*kt.SnmpDeviceConfi
 				removeNum = d.PurgeDevice
 			}
 
-			if removeNum > 0 && pollDuration > 0 {
+			if removeNum == -1 { // If this value is -1, keep this device forever.
+				continue
+			} else if removeNum < -1 {
+				// This is an invalid state, throw an error.
+				return nil, fmt.Errorf("Invalid PurgeDevice value (%d) for device %s", removeNum, d.DeviceName)
+			}
+
+			if removeNum > 0 && pollDuration > 0 { // Keep remove num guard just in case.
 				// Get time this would take.
 				removeTime := time.Now().Add(time.Duration(removeNum) * pollDuration * -1)
 				removeTime = removeTime.Add(RemoveTimeBuffer)
 				if d.Checked.Before(removeTime) {
-					log.Warnf("Removing device %s because it hasn't been seen since %v. Max time %v", d.DeviceName, d.Checked, removeTime)
+					log.Infof("Removing device %s because it hasn't been seen since %v. Max time %v", d.DeviceName, d.Checked, removeTime)
 					delete(conf.Devices, dip)
 					stats.added--
 				}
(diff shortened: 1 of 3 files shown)


jbeveland27:
The changes look good. I do wonder if this is really what we want though, as it's a bit different in intent than what was logged in the issue (this sort of combines both use cases together as a global option). Should we consider making `purge_devices_after_num` per-device?

thezackm:
@araddas - this is based on conversations with you - would you have any feedback on this approach? 

@i3149 / @jbeveland27 - we have a precedent on this type of pattern in the [infra agent config](https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings/#remove-entities-period) already, should we consider modeling after that?

I tend to agree that we want both a global and per-device setting, with the per-device taking priority.

i3149:
Added a per device field `purge_after_num` which lets specific devices get deleted or not. Overrides any global value. 

You can also set this to `purge_after_num: -1` to pin a device and never delete it.

pkg/inputs/snmp/disco.go
```diff
@@ -360,6 +361,27 @@ func addDevices(ctx context.Context, foundDevices map[string]*kt.SnmpDeviceConfi
...
 
+	// Remove any devices which havn't been up long enough if this is set.
+	if !isTest {
+		removeNum := conf.Global.PurgeDevices
+		for dip, d := range conf.Devices {
+			if d.PurgeDevice != 0 { // Let this override the global settings.
+				removeNum = d.PurgeDevice
+			}
+
+			if removeNum > 0 && pollDuration > 0 {
```
jbeveland27, line 372: > You can also set this to purge_after_num: -1 to pin a device and never delete it.

I wonder if we specifically code that behavior here. The way this is now, any non-positive value would technically work to keep a device forever. Just thinking that knowledge will get lost over time unless it's actually in code

jbeveland27 (commented):
Just left 2 small comments. Otherwise looks good

pkg/inputs/snmp/disco.go
```diff
@@ -360,6 +361,27 @@ func addDevices(ctx context.Context, foundDevices map[string]*kt.SnmpDeviceConfi
...
+				// Get time this would take.
+				removeTime := time.Now().Add(time.Duration(removeNum) * pollDuration * -1)
+				removeTime = removeTime.Add(RemoveTimeBuffer)
+				if d.Checked.Before(removeTime) {
+					log.Warnf("Removing device %s because it hasn't been seen since %v. Max time %v", d.DeviceName, d.Checked, removeTime)
```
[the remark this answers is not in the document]
i3149: Good call.

pkg/inputs/snmp/disco.go
i3149: Yeah, good idea to make this this explicit. Now will throw an error if value isn't -1,0,1 to MAX_INT

jbeveland27 (approved):

PR #345, opened 2022-06-29
Add "i" suffix to int field values in influx line

tdanner:
Influx line protocol (specifically https://github.com/influxdata/line-protocol, the go package for parsing it) treats numeric field values as floats unless they have an `i` suffix. As SNMP counters are likely to overflow the $2^{53}$ integer precision of float64, this is important.


i3149 (approved):
LGTM!

Issue #346, opened 2022-06-30
FR Allow single discovery to test v3 and v2 credentials when both are needed


Mesverrum:
For simplicity of administration it would be helpful if users could put snmpv2 and v3 credentials in a single discovery config and have ktranslate test them all.

i3149:
Should be good now.

PR #348, opened 2022-06-30
Allow discovery to use both v3 and v2 snmp creds. 

i3149:
#346

Lets a user mix up v2 and v3 snmp creds in the disco config. All possible creds will be tried, snmpv3 first and then v2.


jbeveland27 (approved):
LGTM 🚢

Issue #347, opened 2022-06-30
FR Add OR capability to filters


hstafford-kentik:
It would be nice if there was a way to combine filters as ORs instead of exclusively ANDs.   For example, only registering flow IN our OUT of a port or a network as SRC or DEST.   

Ideally, some way to combine ANDs and ORs would be ideal, but may be asking too much.

i3149:
Should be good now.

PR #349, opened 2022-06-30
Support for or-ing filters together #347

i3149:
Allow OR'ing filters together like

```
--filters 'int,input_port,==,4 or int,output_port,==,0'
```

This will create 2 filters, which will pass if either one is true. 

Between `--filters` flags, current AND logic applies. So for example:

```
--filters 'int,input_port,==,4 or int,output_port,==,0' --filters 'int,protocol,==,7'
```

Will pass if (input_port = 4 OR output_port = 0) AND (protocol = 7)

Note -- the or string is ` or ` with the 2 spaces being required.

### DIFF
--- a/pkg/cat/types.go
+++ b/pkg/cat/types.go
@@ -96,7 +96,7 @@ type KTranslate struct {
 	rollups      []rollup.Roller
 	doRollups    bool
 	doFilter     bool
-	filters      []filter.Filter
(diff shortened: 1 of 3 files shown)


ehazlett (approved):
LGTM

hstafford-kentik (approved):
Looks good to me from a user level - no comment on the man behind the curtain.

PR #351, opened 2022-07-06
Allow ent mappings for fields 1-5

i3149:
Bumping golfow dependency to 1.1.1 to allow for more enterprise fields.


ehazlett (dismissed):
LGTM

PR #358, opened 2022-07-15
Allow global user tags to be defined on a per provider basis

i3149:
This allows the global tags setting for snmp to be prefixed with a given provider. Only devices who have this provider will get assigned these tags. Currently uses the form `provider:foo:tagName` to prefix these. If the provider is `foo`, then the tag `tagName` will get applied.

### DIFF
--- a/pkg/inputs/snmp/snmp.go
+++ b/pkg/inputs/snmp/snmp.go
@@ -209,11 +209,17 @@ func runSnmpPolling(ctx context.Context, snmpFile string, jchfChan chan []*kt.JC
 			}
 		}
 
-		// Tweak any per provider tags here.
+		// Tweak any per provider tags and match attributes here.
 		for key, tag := range device.UserTags {
 			delete(device.UserTags, key)
 			if nk, ok := matchesPrefix(key, device.Provider); ok {
 				device.UserTags[nk] = tag // We do use this tag, go ahead and update the value to the one returned.
+			}
+		}
+		for key, match := range device.MatchAttr {
+			delete(device.MatchAttr, key)
+			if nk, ok := matchesPrefix(key, device.Provider); ok {
+				device.MatchAttr[nk] = match // We do use this match, go ahead and update the value to the one returned.
 			}
 		}
 
--- a/pkg/inputs/snmp/snmp.go
+++ b/pkg/inputs/snmp/snmp.go
@@ -375,7 +375,7 @@ func parseConfig(ctx context.Context, file string, log logger.ContextL) (*kt.Snm
 	// If there's a global user tags and match, add them in here.
 	if ms.Global != nil {
 		for p, m := range ms.Global.ProviderMap {
-			m.Init(p, ms.Global) // Set up any provider based user and match tags here.
+			m.Init(p, &ms) // Set up any provider based user and match tags here.
 		}
 
 		for k, v := range ms.Global.UserTags {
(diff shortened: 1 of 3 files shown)


i3149:
@araddas if you have any opinions.

araddas:
@i3149 can this be expanded for match attributes as well?  Maybe something like this, with the assumption that if the provider isn't defined in provider_maps then it inherits global values for user_tags and match_attributes.

```
global:
  provider_maps:
    kentik-switch:
      user_tags:
        device_type: switch
        service: core_infra
      match_attributes:
        '!if_Alias': ^Uplink.*
    kentik-firewall:
      user_tags:
        device_type: firewall
        service: edge_infra
      match_attributes:
        '!if_Alias': ^Inside.*
  
```

i3149:
@araddas -- that's great suggestion. Updated the config to support your suggested map above.

pkg/inputs/snmp/snmp.go
```diff
@@ -204,6 +204,20 @@ func runSnmpPolling(ctx context.Context, snmpFile string, jchfChan chan []*kt.JC
 			}
 		}
 
+		// Tweak any per provider tags and match attributes here.
+		for key, tag := range device.UserTags {
+			delete(device.UserTags, key)
```
jbeveland27, line 209: I'm curious about the ordering here. I _think_  this is deleting the existing userTag before we know there's a match from the global `provider_maps`, should that be the case? Does it make sense that there would be userTags set on a device that we don't want to touch? It feels like there's potentially a precedence rule we need to clear up:

```
least specific                           most specific
global        -> provider_tags        -> device user_tags / match_attributes 
```

pkg/inputs/snmp/snmp.go
```diff
@@ -204,6 +204,20 @@ func runSnmpPolling(ctx context.Context, snmpFile string, jchfChan chan []*kt.JC
...
+			if nk, ok := matchesPrefix(key, device.Provider); ok {
+				device.UserTags[nk] = tag // We do use this tag, go ahead and update the value to the one returned.
+			}
+		}
+		for key, match := range device.MatchAttr {
+			delete(device.MatchAttr, key)
```
jbeveland27, line 215: same comment as above

pkg/kt/snmp.go
```diff
@@ -218,20 +220,26 @@ type SnmpDiscoConfig struct {
...
+	MibsEnabled   []string               `yaml:"mibs_enabled"`
+	TimeoutMS     int                    `yaml:"timeout_ms"`
+	Retries       int                    `yaml:"retries"`
+	GlobalV3      *V3SNMPConfig          `yaml:"global_v3"`
+	RunPing       bool                   `yaml:"response_time"`
+	PingSec       int                    `yaml:"ping_interval_sec,omitempty"`
+	PurgeDevices  int                    `yaml:"purge_devices_after_num"` // Delete any device if its not seen after X discovery attempts. Default is 0, which means things never get purged.
+	UserTags      map[string]string      `yaml:"user_tags"`
+	MatchAttr     map[string]string      `yaml:"match_attributes"`
+	ProviderMap   map[string]ProviderMap `yaml:"provider_maps"`
```
jbeveland27, line 242: Nit: Not a big deal, but maybe should name this `providers` instead of `provider_maps` to be consistent with `user_tags` / `match_attributes`

pkg/inputs/snmp/snmp.go
```diff
@@ -371,6 +385,10 @@ func parseConfig(ctx context.Context, file string, log logger.ContextL) (*kt.Snm
 
 	// If there's a global user tags and match, add them in here.
 	if ms.Global != nil {
+		for p, m := range ms.Global.ProviderMap {
```
jbeveland27, line 388: I might just be missing it, but is `ProviderMap` always initialized?

i3149: Its not a pointer so its always (implicitly) set up. I also went around on this for a while but I think this is the state.

pkg/inputs/snmp/snmp.go
i3149: Yar -- you were unfortunately right here. Lots of tests later and I think this does what is lade out above.

jbeveland27 (approved):
LGTM! Thanks for the tests 🥇

PR #362, opened 2022-07-19
Support wildcards in trap oids

i3149:
Addresses https://github.com/kentik/snmp-profiles/pull/365

This allows trap oids like 

```
  - trap_oid: 1.3.6.1.6.3.1.1.5.5.*
    trap_name: authenticationFailure
```

Where any oid with the prefix of 1.3.6.1.6.3.1.1.5.5 will get matched with this trap. 

I think this will allow the linked issue to get closed.

### DIFF
--- a/pkg/inputs/snmp/mibs/load.go
+++ b/pkg/inputs/snmp/mibs/load.go
@@ -77,9 +77,19 @@ func (db *MibDB) Close() {
 }
 
 func (db *MibDB) GetTrap(oid string) *Trap {
-	if t, ok := db.traps[oid]; ok {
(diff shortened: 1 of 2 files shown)


tdanner (dismissed):
lgtm

jbeveland27 (approved):

Mesverrum:
@314 I pinged this on Slack a while back, but we need to  revisit this one.  Once I got in and reviewed the data it looks like i described the problem incorrectly.  The index number is not on the trap_sysoid, those are normal.  It is being attached to the varbind/event fields.  So I think we want to relocate this functionality over from where it is down to the part where we parse those fields.

PR #366, opened 2022-07-21
carbon: add intrinsic and meta tags

ehazlett:
This adds more fields for the carbon formatter.  Intrinsic fields are added for improved dimensions as well as meta tags for more context.


i3149 (approved):
LGTM!

PR #367, opened 2022-07-21
chore/nr snmp base

thezackm:
* adding new minimalist example for `snmp-base.yaml` to the `/deployment` dir to show recommended settings for NR 
* updating NR k8s example to add discovery settings

questions:
* @jbeveland27 - are these the right settings to go forward on?
* @i3149 - will this minimalist config work? several fields that we don't ever change have been redacted, do we need to put them back or update ktranslate to have default values for them by chance?
* @i3149 - please confirm k8s supports discovery. I assume it does but can't recall why the settings weren't in the config example already.

### DIFF
--- a/deployment/docker/snmp-base-nr.yaml
+++ b/deployment/docker/snmp-base-nr.yaml
@@ -1,0 +1,27 @@
+# Recommended default minimalist config for sending data to New Relic
+---
+devices:
+trap:
+  listen: 127.0.0.1:1620
+  community: hello
+discovery:
+  cidrs:
+  - 10.10.0.0/24
+  ignore_list: []
+  debug: false
+  ports:
+  - 161
+  default_communities:
+  - public
+  default_v3: null
+  add_devices: true
+  add_mibs: true
+  threads: 4
+  replace_devices: true
+  check_all_ips: true
+global:
+  poll_time_sec: 300
+  mibs_enabled:
+  - IF-MIB
+  timeout_ms: 3000
+  retries: 0
--- a/deployment/docker/snmp-base-nr.yaml
+++ b/deployment/docker/snmp-base-nr.yaml
@@ -20,6 +20,7 @@ discovery:
   check_all_ips: true
 global:
   poll_time_sec: 300
+  mib_profile_dir: /etc/ktranslate/profiles
   mibs_enabled:
   - IF-MIB
   timeout_ms: 3000
--- a/deployment/docker/snmp-base-nr.yaml
+++ b/deployment/docker/snmp-base-nr.yaml
@@ -3,7 +3,6 @@
 devices:
 trap:
   listen: 127.0.0.1:1620
-  community: hello
 discovery:
   cidrs:
   - 10.10.0.0/24
--- a/deployment/kubernetes/ktranslate-nr.yaml
+++ b/deployment/kubernetes/ktranslate-nr.yaml
@@ -113,7 +113,6 @@ data:
     devices:
     trap:
       listen: 0.0.0.0:1620
-      community: hello
     discovery:
       cidrs:
       - 10.10.0.0/24
--- a/deployment/kubernetes/ktranslate-nr.yaml
+++ b/deployment/kubernetes/ktranslate-nr.yaml
@@ -1,3 +1,6 @@
+# Recommended default config for sending SNMP data to New Relic
+# Includes discovery on startup with a 6 hour recurring interval
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -28,9 +31,11 @@ spec:
               key: nr_account_id
         args:
           - --metalisten=0.0.0.0:8083
-          - --snmp=/etc/ktranslate/snmp.yml
+          - --snmp=/etc/ktranslate/snmp-base.yaml
           - --metrics=jchf
           - --tee_logs=true
+          - --snmp_discovery_on_start=true
+          - --snmp_discovery_min=360
           - nr1.snmp
         resources:
           limits:
@@ -59,8 +64,8 @@ spec:
           periodSeconds: 5
         volumeMounts:
           - name: ktranslate-config
-            mountPath: /etc/ktranslate/snmp.yml
-            subPath: snmp.yml
+            mountPath: /etc/ktranslate/snmp-base.yaml
+            subPath: snmp-base.yaml
       volumes:
         - name: ktranslate-config
           configMap:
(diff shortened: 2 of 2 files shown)


deployment/kubernetes/ktranslate-nr.yaml
```diff
@@ -104,23 +109,28 @@ kind: ConfigMap
...
+  snmp-base.yaml: |
     devices:
-      switch:
-        device_name: switch
-        device_ip: 10.10.0.10
-        flow_only: true
-        user_tags: {}
     trap:
       listen: 0.0.0.0:1620
       community: hello
```
Mesverrum, line 116: If we are in here making changes, the community string on the listener does not appear to do anything, it does not filter incoming messages that use a different community and I have used it successfully without this line in the past.

deployment/docker/snmp-base-nr.yaml
```diff
@@ -0,0 +1,27 @@
+# Recommended default minimalist config for sending data to New Relic
+---
+devices:
+trap:
+  listen: 127.0.0.1:1620
+  community: hello
```
Mesverrum, line 6: ^ Same as my previous comment about not needing a trap listener community

deployment/kubernetes/ktranslate-nr.yaml
thezackm: fair point - @314 - are we clear to remove `trap.community` as well? I think so based on [this bit](https://github.com/kentik/ktranslate/blob/729ff43fb7b949b60f7cfad3a1e474dfeaf80178/pkg/inputs/snmp/traps/traps.go#L62-L63)

```go
	if conf.Trap.Community == "" {
		conf.Trap.Community = "hello"
	}
```

i3149: I think so too. Fine to remove.

deployment/docker/snmp-base-nr.yaml
```diff
@@ -0,0 +1,27 @@
...
+  default_communities:
+  - public
+  default_v3: null
+  add_devices: true
+  add_mibs: true
+  threads: 4
+  replace_devices: true
+  check_all_ips: true
+global:
+  poll_time_sec: 300
```
i3149, line 23: Add this line `mib_profile_dir: /etc/ktranslate/profiles` here.

deployment/kubernetes/ktranslate-nr.yaml
```diff
@@ -104,23 +109,28 @@ kind: ConfigMap
...
+      default_v3: null
+      add_devices: true
+      add_mibs: true
+      threads: 4
+      replace_devices: true
+      check_all_ips: true
     global:
-      poll_time_sec: 30
-      drop_if_outside_poll: false
-      mib_profile_dir: /etc/ktranslate/profiles
```
i3149, line 122: As above, this line is needed to get profiles.

i3149 (commented):
ktrans in k8s does support discovery. Just will need to ensure that the config file is writable by the process. 

Only other thing I see is need for the snmp profiles to be explicit in the config. We may want to look at making this path the default but its not currently.

thezackm:
changes made - pending @jbeveland27 confirmation on the `poll_time_sec` and `check_all_ips` settings being the new standard recc and we can move forward on this one. 

removing the `wip` tag

deployment/kubernetes/ktranslate-nr.yaml
```diff
@@ -104,23 +109,28 @@ kind: ConfigMap
...
+      - 161
+      default_communities:
+      - public
+      default_v3: null
+      add_devices: true
+      add_mibs: true
+      threads: 4
+      replace_devices: true
+      check_all_ips: true
     global:
```
jbeveland27, line 131: The [wiki](https://github.com/kentik/ktranslate/wiki/Sample-SNMP-config#global-config-section) indicates `global.retries` is a required field. Is that accurate?

jbeveland27: Maybe `mib_profile_dir` should be required in the wiki as well 🤔

deployment/kubernetes/ktranslate-nr.yaml
```diff
@@ -28,9 +31,11 @@ spec:
               key: nr_account_id
         args:
           - --metalisten=0.0.0.0:8083
-          - --snmp=/etc/ktranslate/snmp.yml
+          - --snmp=/etc/ktranslate/snmp-base.yaml
           - --metrics=jchf
           - --tee_logs=true
+          - --snmp_discovery_on_start=true
+          - --snmp_discovery_min=360
```
jbeveland27, line 38: I was (somewhat arbitrarily) thinking 1 hour, but 6 is good. We can adapt our UI to do the same

deployment/docker/snmp-base-nr.yaml
```diff
@@ -0,0 +1,27 @@
+# Recommended default minimalist config for sending data to New Relic
+---
+devices:
+trap:
+  listen: 127.0.0.1:1620
+discovery:
+  cidrs:
+  - 10.10.0.0/24
```
jbeveland27, line 8: Is there a reason we default to `10.10.0.0/24`? 

I mentioned this before and think it's worth bringing up again: would it be useful to have ktranslate use `<host_ip>/24` as a default, and use `{P,p}ublic / {P,p}rivate` as the community strings? Reason being - this might be a sensible default to just get a user going quickly and it'd discover reachable devices attached to that host. It wouldn't be the true production deployment, but at that point users are likely getting more involved in the configuration anyways. We could still have input fields in the UI so the user can edit CIDRs and community strings, but if they omit then we can fall back to these defaults.

We'd just have to figure out some way to indicate that behavior in the config. I floated this idea by @i3149 a few weeks ago and he indicated it'd be feasible and might make sense as a sensible default. Thoughts?

Mesverrum: I'll throw in that as a general rule I have been strongly discouraging anyone from using ktrans to monitor servers because net-snmp is such a huge downgrade compared to our agent.   Making it the default entity undermines that a bit, but I think the value in just being able to discover a single local thing without relying on other teams for firewalls and access and such is worth it.

i3149: `10.10.0.0/24` Is how my home network is set up. No better reason that this for why the default.

deployment/kubernetes/ktranslate-nr.yaml
i3149: Good call out on retries. 0 retries is the default which seems non optimal -- that's why its required, but things will work with 0 retries.

deployment/docker/snmp-base-nr.yaml
thezackm: > but I think the value in just being able to discover a single local thing without relying on other teams for firewalls and access and such is worth it.

I think there's a discussion to be had around this. IMO there should be a purposeful drive to installation of this agent and opening up paths towards monitoring entities following anti-patterns just to show success with ktranslate seems incorrect. I'm trying to think of the value as a user of seeing ktranslate able to monitor a server via SNMP vs a better experience that guides me towards getting my firewall/access fixed.

jbeveland27: Agreed, a discussion would be good. I don't think we'd want to drive monitoring of a server via SNMP (we'd push the infra agent for that use case). I was more so wondering if it's feasible to reach even the first network hop with some sensible defaults. Get that first device into NR and then figure out how to get more. But if it's not worth the time/effort + will be error prone, then we can just ignore my musings and go back to regularly scheduled programming 😄

jbeveland27 (approved):

PR #370, opened 2022-07-29
Update snmp trap listen IP

jchurchmanNR:
Per Marc, to coincide with other minor setup changes from the New Relic setup flow.

### DIFF
--- a/config/snmp-base.yaml
+++ b/config/snmp-base.yaml
@@ -1,6 +1,6 @@
 devices:
 trap:
-  listen: 127.0.0.1:1620
+  listen: 0.0.0.0:1620
   community: hello
   version: ""
   transport: ""


i3149 (commented):
I'm leary about doing this as the default because its a potential security risk. I'd prefer to force the user to explicitly turn trap listening on for all interfaces, vs defaulting to this. We also don't default listen anywhere else in the code base.

jchurchmanNR:
> I'm leary about doing this as the default because its a potential security risk. I'd prefer to force the user to explicitly turn trap listening on for all interfaces, vs defaulting to this. We also don't default listen anywhere else in the code base.

Understandable. I'm going to defer to @Mesverrum (or maybe @thezackm ) to engage regarding this concern.

jchurchmanNR:
@i3149 Spoke further with @Mesverrum and am pursuing an alternate solution. Thank you

PR #371, opened 2022-07-29
String-valued fields must be "quoted" and escaped differently than tags

tdanner:
The influx parser throws errors like "at line 168:251: field value has unrecognized type" when a field's value is an unquoted string.

### DIFF
--- a/pkg/formats/influx/influx.go
+++ b/pkg/formats/influx/influx.go
@@ -43,7 +43,7 @@ func (d *InfluxData) GetTags() string {
 		switch t := v.(type) {
 		case string:
 			if t != "" {
-				tags = append(tags, fmt.Sprintf("%s=%s", kval, influxEscape(t)))
+				tags = append(tags, fmt.Sprintf("%s=%s", kval, influxEscapeTag(t)))
 			}
 		default:
 			tags = append(tags, fmt.Sprintf("%s=%v", kval, v))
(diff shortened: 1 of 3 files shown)


pkg/formats/influx/influx.go
```diff
@@ -66,7 +66,7 @@ func (d *InfluxData) String() string {
 	i := 0
 	for k, v := range d.Fields {
 		if ev, ok := d.Tags[k]; ok { // There's an enum here, use this vs the int value.
-			fields[i] = k + "=" + ev.(string)
+			fields[i] = k + "=\"" + influxEscapeField(ev.(string)) + "\""
```
i3149, line 69: you could write this as

```
fields[i] = fmt.Sprintf("%s=%q", k, influxEscapeField(ev.(string)))
```
Relying on %q to properly quote the value. 

Also, will ev ever not be a string?

i3149 (approved):
lgtm, one note on quoting things.

pkg/formats/influx/influx.go
tdanner: `%q` will wrap it in `"` and also apply escaping to make the result a valid Go string literal. That would do the right thing with internal `"` and `\` (good), but I think it would also turn newlines into `\n` and probably some other special characters. This probably doesn't matter in this code path since these are only enum names coming from snmp profile yaml files, but I'd rather keep it explicit.

PR #372, opened 2022-08-02
Put \n at the end of every influx line, not just between them

tdanner:
This is a cherry-pick of https://github.com/kentik/ktranslate/pull/340/commits/10e11361137865a47e4a856ea315660eabefb05c, which I made and then apparently accidentally overwrote in the same PR.

I came back to this because I was seeing joined-up influx lines in the _file_ output from ktranslate.

### DIFF
--- a/pkg/formats/influx/influx.go
+++ b/pkg/formats/influx/influx.go
@@ -1,6 +1,7 @@
 package influx
 
 import (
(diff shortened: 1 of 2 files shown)


i3149 (approved):
Right -- makes sense.

PR #376, opened 2022-08-04
Influx: use github.com/influxdata/line-protocol/v2 Encoder instead of strings

tdanner:
I am still hitting edge cases with influx encoding. Rather than continue to read https://docs.influxdata.com/influxdb/cloud/reference/syntax/line-protocol/ ever more closely and keep tweaking the string handling in influx.go, I think it would be better to switch to the "official" encoder.

Right now I have it checking the error state of the encoder after each step and panicking if it is non-nil. This is quite verbose and, while helpful during debugging, may not be how we want to handle errors in production. Feedback welcome.

### DIFF
--- a/pkg/formats/influx/influx.go
+++ b/pkg/formats/influx/influx.go
@@ -1,16 +1,14 @@
 package influx
 
 import (
-	"bytes"
 	"flag"
 	"fmt"
-	"regexp"
 	"sort"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/influxdata/line-protocol/v2/lineprotocol"
 	"github.com/kentik/ktranslate/pkg/formats/util"
 	"github.com/kentik/ktranslate/pkg/kt"
 	"github.com/kentik/ktranslate/pkg/rollup"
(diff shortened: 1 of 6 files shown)


i3149 (commented):
I'm personally not a fan of panic statements in agent code. Since its so hard to get customers to run it/update it, better to make agents as bulletproof as possible. Rather than panicking on each error, I would consider logging an error message once and then updating an internal metric with the error count. 

Downside of this is that we're not currently passing the metrics registry into the format block, but this should be easy enough to do.

tdanner:
Sounds good - I'll plumb the telemetry stuff into formatters and log errors instead of calling panic.

pkg/formats/influx/influx.go
```diff
@@ -117,7 +91,16 @@ func NewMergedInfluxData(s InfluxDataSet) *InfluxData {
 
 type InfluxDataSet []InfluxData
 
-func (s InfluxDataSet) Bytes() []byte {
+func (f *InfluxFormat) report(err error, format string, args ...interface{}) {
+	str := fmt.Sprintf(format, args...)
+	if !f.invalids[str] {
```
i3149, line 96: There's a small chance this will panic with a concurrent map access. Consider wrapping into a sync.RWMutex?

i3149 (approved):
Looks awesome! Probably worth adding a mutex just because if there's a ton of traffic which is also erroring you'll probably hit a panic.

pkg/formats/influx/influx.go
tdanner: Good point. Will do.

i3149 (approved):
Ship it!

PR #379, opened 2022-08-05
Config file Support

ehazlett:
This adds support for configuring ktranslate via config file by using a new `--config` flag. You can also generate a default configuration with all available options by using `ktranslate --generate-config`. All current flags _should_ still be active and supported. The config file format is `yaml` and looks like this:

```yaml
listenaddr: 127.0.0.1:8081
dns: "1.1.1.1"
processingthreads: 1
inputthreads: 1
maxthreads: 1
format: flat_json
compression: none
sinks: stdout
maxflowspermessage: 10000
rollupinterval: 0
rollupandalpha: false
samplerate: 1
samplemin: 1
enablesnmpdiscovery: false
kentikemail: "[EMAIL]"
kentikapitoken: "[TOKEN]"
kentikplan: [PLAN_ID]
apibaseurl: https://api.kentik.com
enableteelogs: false
enablehttpinput: true
server:
    servicename: "ktranslate"
    loglevel: info
    logtostdout: true
```

There is a full example with all available options in `config.yaml.example`. While not recommended it should be fine to mix and match the config file with flags where the flags take precedence. The deploy configs in `hack/` have also been updated to show examples on usage.


i3149 (approved):
LGTM!

i3149 (approved):
:rocket: :rocket:

PR #380, opened 2022-08-05
Allowing discovery on start without looping

i3149:
Now you can set the flag `-snmp_discovery_on_start` separately without also setting `-snmp_discovery_min`. 

Also fixes a bug where if a new device is discovered on startup discovery actually happens twice.


thezackm (dismissed):

ehazlett (approved):
LGTM

PR #382, opened 2022-08-09
Remove chars 0-0x1F and 0x7F from tag values

tdanner:
I was hoping that InfluxDB's influx encoder would take care of all character encoding issues, but I should have actually read the code first. If you give it a tag value with an "invalid" character, it doesn't escape it or even remove it - it just flags it as an error and outputs nothing for that whole line.

So we need to filter tag values to remove disallowed characters - 0x00-0x1F (which includes \n and \r) plus 0x7F - before sending the tag value to the encoder.

Where this is coming up: Cisco devices have multiline strings for SysDescr.

### DIFF
--- a/pkg/formats/influx/influx.go
+++ b/pkg/formats/influx/influx.go
@@ -1,6 +1,7 @@
 package influx
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"sort"
@@ -148,7 +149,7 @@ line:
 					s = fmt.Sprintf("%v", v)
 				}
 				if s != "" {
-					enc.AddTag(k, s)
+					enc.AddTag(k, prepareTagValue(s))
 				}
 				if enc.Err() != nil {
 					f.report(enc.Err(), "tag '%s'='%v'", k, s)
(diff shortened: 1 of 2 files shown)


i3149 (commented):
What do you think about the Map function vs ContainsAny?

```
	swap := func(r rune) rune {
		switch {
		case r >= 0x00 && r <= 0x1f:
			return ' '
		case r == 0x7f:
			return ' '
		}
		return r
	}
	return strings.Map(swap, s)
```

I find this logic a little easier to follow than yours. I haven't tested perf though.

tdanner:
The map method is a little bit slower for clean strings (no characters need replacing) and a lot faster for dirty strings. If you check for the presence of invalid characters first, it is just as fast for clean strings (since it is the same code for that path) and still much faster for dirty strings, though not as fast as if you just call map first.

Without the pre-check:

```
❯ go test -bench=.
goos: darwin
goarch: arm64
pkg: github.com/kentik/ktranslate/pkg/formats/influx
BenchmarkPrepareTagValueContainsAnyClean-10     13377864                87.39 ns/op
BenchmarkPrepareTagValueContainsAnyDirty-10      2344202               507.9 ns/op
BenchmarkPrepareTagValueMapClean-10             11286318               105.2 ns/op
BenchmarkPrepareTagValueMapDirty-10              7844433               151.4 ns/op
```

With it:

```
❯ go test -bench=.
goos: darwin
goarch: arm64
pkg: github.com/kentik/ktranslate/pkg/formats/influx
BenchmarkPrepareTagValueContainsAnyClean-10     12530122                86.41 ns/op
BenchmarkPrepareTagValueContainsAnyDirty-10      2357702               507.6 ns/op
BenchmarkPrepareTagValueMapClean-10             13497147                86.24 ns/op
BenchmarkPrepareTagValueMapDirty-10              5600560               213.7 ns/op
```

i3149 (approved):
LGTM then. Assume the clean case is dominating so makes sense to optimize here.

PR #385, opened 2022-08-10
Fix multiple header parse issue with config

ehazlett:
This fixes an issue where the config parsing didn't detect multiple `http_header` flags.


tdanner (approved):
Tested locally and found working.

PR #388, opened 2022-08-16
Moving if utilization % to look like other if metrics in influx

i3149:
This should cover

```
the second one comes from snmp-profile if-mib.yaml which is good
[@314](https://kentik.slack.com/team/U06CAJBJR)
 and 
[@tdanner](https://kentik.slack.com/team/U037X8HS6K0)
, I understand that utilization metrics are additionally calculated as they are not in the snmp-profile specification, but does it make sense to glue these together with the rest of the interface metrics and having just one record instead of two? I guess its better if we do it at the ktranslate level, than on kmetrics ingest, but its your call? (edited)
```


tdanner (approved):
:shipit:

PR #393, opened 2022-08-18
Remove leading `.` from SNMP indexes (fixes #390)

### LINKED ISSUE #390: remove leading "." from the "index" dimension
remove leading "." from the "index" dimension

tdanner:
Note that doing it this way affects all outputs, not just influx. If this is a problem, I could look at doing it just for influx. However, I prefer the no-leading-dot style and would rather apply it universally if that is not too disruptive.


i3149 (approved):
I think its fine to do this globally.

PR #394, opened 2022-08-18
"removeAttrForSnmp" filter only for NR, not influx format (fixes #392)

### LINKED ISSUE #392: List of tags to remove for NR should be applied only for NR output
in util.go, there is a list of tags that should be removed for NR. This should be only applied in the case of NR output, but not for other outputs, like InfluxDB. 

```
// List of attributes to not pass to NR.
var removeAttrForSnmp = []string{
	"Uptime",
	"if_LastChange",
	"SysServices",
	"if_Mtu",
	"if_ConnectorPresent",
	"output_port",
	"input_port",
	"DropMetric",
	"sysoid_vendor",
}
```

tdanner:


i3149 (approved):
I like it.

PR #395, opened 2022-08-18
Only stripping metrics from attributes in influx

i3149:
Actually #391 breaks New Relic. This puts a guard in so that only influx will remove string metrics as tags.


tdanner (approved):
Tricky stuff. We might need a more sophisticated integration test system to keep us from breaking and rebreaking these filter behaviors.

i3149:
Agreed, the tests really need to be beefed up. Lots of room for regressions here.

PR #400, opened 2022-08-22
Ignoring 0 byte flow packets now

i3149:
Incorporates https://github.com/netsampler/goflow2/pull/114

This fixes log errors when 0 byte packets are mixed into valid flow.


jbeveland27 (approved):
LGTM

PR #405, opened 2022-08-25
Allow same key multiple tables

i3149:
This one goes with https://github.com/kentik/snmp-profiles/pull/411

It allows the same tag to be used across mibs (when their tables line up). I think this makes sense and will not cause cross table junking up of attributes.


jbeveland27 (approved):
LGTM

dpajin-kentik:
I can test when its merged :)

PR #408, opened 2022-08-26
Trying a reverse dns lookup if trap sender is not known

### LINKED ISSUE #398: FR: Use DNS to resolve IP addresses
A customer currently has to "hard-configurate" the domain-name of a device with a specific IP address. In bigger environments, this will be handled by a DNS, which makes it impossible to keep telemetry up-to-date if an IP address changes. 

The requirement is to be able to define a DNS that is being used to lookup the device-domain based on the IP address, both for SNMP queries and for traps.

i3149:
Fixes #398 
Passes in the dns resolver to ktrans trap handler and will use this if the IP of a trap sender is not known. 

Make sure to also set `--dns=local` as a flag.

### DIFF
--- a/pkg/cat/kkc.go
+++ b/pkg/cat/kkc.go
@@ -658,7 +658,7 @@ func (kc *KTranslate) Run(ctx context.Context) error {
 		}
 		assureInput()
 		kc.metrics.SnmpDeviceData = kt.NewSnmpMetricSet(kc.registry)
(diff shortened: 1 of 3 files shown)


jbeveland27 (approved):

sskilbred:
Hey - this seems to be blocked only by the authorisation level of @i3149 , correct? Is this something that can be solved asap? Asking since I am waiting for a solution for https://github.com/kentik/ktranslate/issues/398 :) Cheers

Issue #407, opened 2022-08-26
💡 FR - Add additional sources for Flow data


thezackm:
As a user, I would like additional out-of-box support for standard flow patterns from Cisco ASA and Cisco NBAR2. This will allow me to easily process these records without needing to provide custom field mapping details at runtime. 

Proposed solution: add 2 new options for flow sources:

```
nf.source=asa
nf.source=nbar2
```

PR #411, opened 2022-08-30
Adding asa,nbar,pan ent field support as native flow

i3149:
This is a way to get extended IPFIX fields from a few common senders: ASA, NBAR and Palo Alto.

The flag `-nf.source` is extended with 3 new values: `asa,nbar,pan`. 

These will accept specific flow from a Cisco ASA, Nbar and Palo Alto firewall. See the special.go file for how these fields are mapped in. Eliminates the need for a config json file to get passed in for every one of these boxes.

### DIFF
--- a/cmd/ktranslate/main.go
+++ b/cmd/ktranslate/main.go
@@ -90,7 +90,7 @@ func init() {
 	flag.StringVar(&sslKeyFile, "ssl_key_file", "", "SSL Key file to use for serving HTTPS traffic")
 	flag.StringVar(&tagMapType, "tag_map_type", "", "type of mapping to use for tag values. file|null")
 	flag.StringVar(&vpcSource, "vpc", kt.LookupEnvString("KENTIK_VPC", ""), "Run VPC Flow Ingest")
-	flag.StringVar(&flowSource, "nf.source", "", "Run NetFlow Ingest Directly. Valid values here are netflow5|netflow9|ipfix|sflow")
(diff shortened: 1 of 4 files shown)


i3149:
Closes #407

thezackm (approved):
@i3149 - is there any way to add a catch-all for IPFIX that basically adds `custom_field_n` for any non-mapped fields that are collected? thinking about future state and potential for custom mappings that we haven't run across yet. 

e.g.; if someone had custom fields called `city` and `state` that they sent in with their IPFIX, instead of just dropping data, we could do something like this:

```
in_bytes: 123,
out_bytes: 456,
...
custom_field_1: "las vegas",
custom_field_2: "NV"
```

_(`city` and `state` aren't the best examples, but think of what we recently ran across with VPCs exported through a Cisco appliance for instance)_

thoughts?

i3149:
@thezackm - its not super possible to look at all fields sent because of how the goflow2 library works. The model it has is explicit opt in and anything not opted in is dropped before the top layer application can see anything. This makes it work well with the static protobuf it uses for passing data around internally but also limits things like you call out. 

I made one change though with nbar, mapping `application` now to the current application field and not overriding it with port based app if set. This should allow existing dashboards to pick up and use `application` without modification for nbar speaking devices.

thezackm (approved):

jbeveland27 (approved):
LGTM 🍰

PR #417, opened 2022-08-30
allow simple listen with /service_info

### LINKED ISSUE #412: FR: Create an independent healthcheck/status endpoint for agent health
When performing a scripted install of ktranslate, it's helpful to utilize the `/healthcheck` endpoint that's exposed as part of the `metalisten` subsystem. This `healthcheck` serves as a good indicator for ensuring ktranslate has been installed successfully. However, the `metalisten` subsystem is more of a developer feature, and turning that on for all customers could have some unintended side effects (particularly around security). 

The proposal is to create a standalone healthcheck endpoint/API that's always on and bound to a known port (port could be configurable via the global config map).

Here's a similar idea that's present in the NR infra agent for inspiration: https://github.com/newrelic/infrastructure-agent/blob/d6890477683c771c774c857cbd9d6d4317bae12e/docs/status_api.md

i3149:
What do you think about this? 

Piggybacks on the existing `listen` handler to have a very basic hc port here too. Defaults to `127.0.0.1:8081` unless otherwise specified. 

Closes #412.

For example:

```
curl 127.0.0.1:8081/service_info | jq .
{
  "Flows": 0,
  "DroppedFlows": 0,
  "FlowsOut": 0,
  "Errors": 0,
  "AlphaQ": 0,
  "JCHFQ": 8000,
  "AlphaQDrop": 0,
  "InputQ": 0,
  "InputQLen": 0,
  "OutputQLen": 0,
  "Sinks": {
    "stdout": {}
  },
  "SnmpDeviceData": {},
  "Inputs": {}
}
```

### DIFF
--- a/pkg/cat/kkc.go
+++ b/pkg/cat/kkc.go
@@ -539,6 +539,7 @@ func (kc *KTranslate) reduce(in []*kt.JCHF) []*kt.JCHF {
 func (kc *KTranslate) getRouter() http.Handler {
 	r := kmux.NewRouter()
 	r.HandleFunc(HttpAlertInboundPath, kc.handleFlow)
+	r.HandleFunc(HttpInfoPath, kc.HttpInfo)
 	r.HandleFunc(HttpHealthCheckPath, func(w http.ResponseWriter, r *http.Request) {
 		fmt.Fprintf(w, "OK\n") // nolint: errcheck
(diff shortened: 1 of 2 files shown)


jbeveland27 (approved):
LGTM 🚢

PR #419, opened 2022-09-01
Fix issue with influxdb formatter

ehazlett:
This fixes an issue with the influxdb formatter where the measurement prefix is empty.

### DIFF
--- a/hack/influxdb/docker-compose.yml
+++ b/hack/influxdb/docker-compose.yml
@@ -35,7 +35,7 @@ services:
         maxthreads: 1
         format: "influx"
         compression: "none"
-        sinks: "http"
+        sinks: "stdout"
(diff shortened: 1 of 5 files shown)


hack/influxdb/docker-compose.yml
```diff
@@ -20,7 +20,7 @@ services:
       - 8086:8086
 
   ktranslate:
-    image: kentik/ktranslate:latest
+    image: docker.io/ehazlett/ktranslate:dev
```
tdanner, line 23: Probably shouldn't merge this part.

ehazlett: It's a test image but ya good catch.

tdanner (approved):

PR #427, opened 2022-09-07
Write out an id file with ed25519 key if id_file flag is set. #415

### LINKED ISSUE #415: FR: Generate a persistent and unique identifier whenever ktranslate is installed
Related to #413 / #414 - we need to uniquely identify each running instance of ktranslate to identify a saved config when querying/mutation from NerdGraph.

This UUID will need to be persisted somewhere locally and remain the same between restarts and upgrades of ktranslate. For containerized deployment we could use a volume. For packaged installs, perhaps the `/etc/default/ktranslate.env` file? Open to suggestions, but just want to emphasize that the solution needs to be durable to avoid losing the config and ending up in an unexpected state.

i3149:
This adds the flag and config option `id_file` / `IDFileLocation`

If set, a ed25519 pub/private keypair will get written out to the specified location. On reload of the program, a short signed message also in this file will get verified. Not doing anything with the id yet, but it should be persisted across runs. 

Note with docker / k8s its up to the user to do the right thing and set this file somewhere which will persist.  

Closes #415.

### DIFF
--- a/cmd/ktranslate/main.go
+++ b/cmd/ktranslate/main.go
@@ -560,6 +560,8 @@ func applyFlags(cfg *ktranslate.Config) error {
 				cfg.Server.OllyDataset = val
 			case "olly_write_key":
 				cfg.Server.OllyWriteKey = val
+			case "id_file":
+				cfg.Server.IDFileLocation = val
 			// pkg/api
 			case "api_device_file":
 				cfg.API.DeviceFile = val
--- a/config.go
+++ b/config.go
@@ -124,13 +124,13 @@ type ServerConfig struct {
 	privKey         []byte
 }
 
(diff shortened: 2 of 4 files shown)


jbeveland27 (commented):
Code-wise I think this all makes sense. I want to take a closer look at this next week though. 

I also had some thought that it might make sense to generate an ID from our backend and have ktranslate call for new/existing IDs via API by passing in some predetermined info (hostname/nr_account_id/etc type info that might let us generate unique IDs in a deterministic manner). However this approach might work and might be more durable as well. I'll verify some things next week and report back.

ehazlett:
Is there a reason why not to just use a UUID?  Is it to validate the ID or perhaps a future config?

PR #430, opened 2022-09-09
WIP -- look at allowing regex to be named fields #402

### LINKED ISSUE #402: FR: allow multiple tags using the same OID
I would like to be able to create multiple tags which are using the same OID.
Currently it seems that this is not possible and it is not completely clear to me why.
Here is the example of what I am trying to do:
```
  - tag: os_name
    column:
      OID: 1.3.6.1.2.1.1.1.0
      name: sysDescr
      regexp: kernel\s+(\w+)\s+[\w\.\-]+,
  - tag: os_version
    column:
      OID: 1.3.6.1.2.1.1.1.0
      name: sysDescr  
      regexp: kernel\s+\w+\s+([\w\.\-]+),
```
even if I put the unique "name", instead of sysDescr, this will not work as it might be overlapping with the sysDescr in system-mib or with something which is hardcoded....

i3149:
What do you guys think about this? 

Rather than allowing duplicate oids which would be a real nightmare to support, this instead allows named [regexes](https://github.com/StefanSchroeder/Golang-Regex-Tutorial/blob/master/01-chapter2.markdown#named-matches) to be matched on multiple submatches. For example:

```
      - column:
          OID: .1.0.8802.1.1.2.1.3.4.0
          name: extendedDesc
          conversion: "regexp:Juniper Networks, Inc. (?P<os_name>\\S+?) , version (?P<os_version>\\S+)"
```

Results in

```
kentik.flowdevice,SysDescr=Juniper\ Networks\,\ Inc.\ mx80-48t\ internet\ router\,\ kernel\ JUNOS\ 13.3R9.13\,\ Build\ date:\ 2016-03-01\ 09:06:52\ UTC\ Copyright\ (c)\ 1996-2016\ Juniper\ Networks\,\ Inc.,SysLocation=iad1-dc3-5012,SysName=edge01.iad1,SysObjectID=.1.3.6.1.4.1.2636.1.1.1.2.57,device_ip=209.50.158.253,device_name=edge01.iad1,instrumentation.name=juniper-mx-router,os_name=mx80-48t,os_version=13.3R9.13,src_addr=209.50.158.253 Uptime=2036791728i 1662686373000000000
```

Pulling out the fields `os_name=mx80-48t` and `os_version=13.3R9.13` via the single entry of `.1.0.8802.1.1.2.1.3.4.0`. 

Closes #402

### DIFF
--- a/pkg/inputs/snmp/metadata/device_metadata.go
+++ b/pkg/inputs/snmp/metadata/device_metadata.go
@@ -158,7 +158,7 @@ func (dm *DeviceMetadata) poll(ctx context.Context, server *gosnmp.GoSNMP) (*kt.
 		case SNMP_sysServices:
 			md.SysServices = int(snmp_util.ToInt64(value))
 		case SNMP_engineID:
-			_, md.EngineID = snmp_util.GetFromConv(wrapper.variable, snmp_util.CONV_ENGINE_ID, dm.log)
+			_, md.EngineID, _ = snmp_util.GetFromConv(wrapper.variable, snmp_util.CONV_ENGINE_ID, dm.log)
 		default:
 			// Now we're actually in the range of custom fields.
 			if wrapper.mib == nil { // This should never happen here.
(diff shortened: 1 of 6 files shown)


dpajin-kentik:
Looks good, it resolves this use case, so it make sense to proceed with it ... and if we hit the wall somewhere else with unique OIDs, we will figure it out later how to overcome it. Thanks

dpajin-kentik (approved):

PR #433, opened 2022-09-12
Fixing gap with bad snmpv3 server

i3149:
I found a case where a bad snmp server can trick the client into thinking there's a working v3 connection when actually it should be trying v2. This adds another guard to keep the disco discoing.


jbeveland27 (approved):
> keep the disco discoing

🕺🏼

PR #438, opened 2022-09-14
Testing out trying to multiplex flow across multiple nr accounts

i3149:
This is a hack to allow multiple new relic accounts to be fed from one firehose instance. Makes life a lot easier for sending data from one process inside kentik to NR.

### DIFF
--- a/config.go
+++ b/config.go
@@ -217,7 +217,7 @@ type Config struct {
 	FormatRollup        string
 	FormatMetric        string
 	Compression         string
-	Sinks               string
+	Sinks               []string
 	MaxFlowsPerMessage  int
 	RollupInterval      int
 	RollupAndAlpha      bool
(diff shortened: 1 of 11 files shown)


config.go
```diff
@@ -288,15 +305,14 @@ func DefaultConfig() *Config {
 		Format:              "flat_json",
 		FormatRollup:        "",
 		Compression:         "none",
-		Sinks:               "stdout",
+		Sinks:               []string{"stdout"},
```
ehazlett, line 308: Will this break current deployments with the comma separated string?

ehazlett: disregard -- i didn't see it in the adapter

i3149: yeah -- I think the split in cmd/ktransate should take care of backwards compatibility.

ehazlett (approved):

PR #444, opened 2022-09-29
Test of games to play with starlark

i3149:
Here's something I'm screwing around with for starlark. What do you guys think? It basically lets you pass a script into the code using the `-enricher` flag and then will run all the found messages through this script. Script can mutate the messages however it likes. Needs a lot of work but this is the basic idea. Focusing on working via a test now, haven't tried this out in the real world.


i3149:
OK -- what do you guys think? This supports 2 main ways to add a script in. 

1) Add in a `script` field for a snmp profile:

```
      # Synology disk type
      - column:
          name: diskType
          OID: 1.3.6.1.4.1.6574.2.1.1.4
          script: |

            def main(n):
              if n.foo != "":
                n.XXXMyIndex = n.foo + "this is a test"
              else:
                n.XXXMyIndex = "missing"
              return None
```

This script will be evaluated only for this given attribute. 

Or: 

```
	EnricherSource      string
	EnricherScript      string
```

Put one of these into the config.yaml file. Then it will be evaluated for each message which goes through ktranslate like any other enrichment. See an example in the test.

ehazlett:
I think this OK to add to the enricher. I think something more generic like https://github.com/ehazlett/ktranslate/tree/processors might be something we can look at in the future but this is cool to start with.

ehazlett (approved):

PR #445, opened 2022-10-03
Make Kentik / KFlow a normal sink

i3149:
Two parts here:

1) Strips out any special handling for kflow and makes kentik a normal sink like any other. Also puts together the kflow format so that it can be sent back into kentik.

2) Adds an enricher option which hashes ips. Turn this on with `-enricher hash_dst_ip12345` flag, where this will hash the destination ips and add the salt 12345. Hash src with `hash_dst_ip` and both with `hash_ip`.

### DIFF
--- a/pkg/cat/kkc.go
+++ b/pkg/cat/kkc.go
@@ -385,9 +385,9 @@ func (kc *KTranslate) sendToSinks(ctx context.Context) error {
 }
 
 // This processes data from the non-kentik input sets.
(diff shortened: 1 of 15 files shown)


pkg/cat/kkc.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/kentik/ktranslate/pkg/maps"
 	"github.com/kentik/ktranslate/pkg/rollup"
 	ss "github.com/kentik/ktranslate/pkg/sinks"
-	"github.com/kentik/ktranslate/pkg/sinks/kentik"
+	//"github.com/kentik/ktranslate/pkg/sinks/kentik"
```
ehazlett, line 28: can we remove this?

ehazlett (approved):

pkg/cat/kkc.go
i3149: good catch.

ehazlett (approved):

PR #446, opened 2022-10-04
Fixing a bug where some egress flow is attributed to ingess values

i3149:
```
background: he originally setup netflow v9 and was only receiving ingress (in_bytes | in_packets); through some digging we found this is by design with pfSense. :facepalm:
now he’s updated to IPFIX and we’ve got a .pcap showing all of the correct template fields, but he’s still only seeing the ingress side of his flows.
details attached here. open to any ideas you may have.
```

This will 1) Set a `FlowDirection` field for netflow.
2) Correctly set `out_bytes` and `out_pkts` for egress flow.

### DIFF
--- a/config.go
+++ b/config.go
@@ -8,9 +8,9 @@ import (
 
 const (
 	// FlowFields are fields for the Flow input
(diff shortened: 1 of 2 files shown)


thezackm (approved):
LGTM

PR #448, opened 2022-10-05
changing name of cell provider 

i3149:
`provider` is a reserved attribute name. This changes the cell provider in meraki to avoid stomping on this name.


thezackm (approved):
🚀

PR #450, opened 2022-10-05
Support or a trap-only value in ktranslate

### LINKED ISSUE #447: FR: Configure kTranslate container/process as trap_only
The request is to add a configuration option that specifies a container or process as trap_only (similar to the flow_only setting). 
The provider attribute added to the New Relic trap event (KSnmpTrap) should be 'kentik-trap-device'.
If the setting is specified on the container, all trap events should be decorated with the mentioned provider attribute value.

The feature will allow New Relic to create entities based on trap-events, a requirement that has been communicated by customers. 

image

i3149:
Closes #447 @sskilbred 

Adds an option `trap_only` to the snmp.conf file. For example:

```
trap:
  listen: 127.0.0.1:1620
  community: ""
  version: v2c
  transport: ""
  v3_config: null
  trap_only: true
```

When true, no snmp polling will happen, but all traps will be handled. Additionally, the provider will be `kentik-trap-device`.

### DIFF
--- a/pkg/inputs/snmp/snmp.go
+++ b/pkg/inputs/snmp/snmp.go
@@ -95,7 +95,9 @@ func StartSNMPPolls(ctx context.Context, jchfChan chan []*kt.JCHF, metrics *kt.S
 	}
 
 	// Now, launch a metadata and metrics server for each configured or discovered device.
-	go wrapSnmpPolling(ctx, snmpFile, jchfChan, metrics, registry, apic, log, 0, cfg)
+	if conf.Trap != nil && !conf.Trap.TrapOnly { // Unless we are turning off everything but snmp traps.
+		go wrapSnmpPolling(ctx, snmpFile, jchfChan, metrics, registry, apic, log, 0, cfg)
+	}
 
 	// Run a trap listener?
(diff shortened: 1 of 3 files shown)


ehazlett (approved):

sskilbred:
Hi - do you have an ETA on when this will be merged? We want to set up and test the synthesis process on New Relic side. Cheers

i3149:
@jbeveland27  -- does this look good to you?

jbeveland27 (approved):
LGTM

jbeveland27:
@i3149 This totally works, but I wonder if the approach should've been a flag vs an implicit behavior embedded within the traps config. `-traps_only=true` is more aligned with the existing `-flow_only=true` approach for flows.

@ehazlett Should this also include an update to the [SnmpInputConfig](https://github.com/kentik/ktranslate/blob/main/config.go#L151) to support a `TrapsOnly`? Also makes me wonder if something needs to change to avoid drift in capabilities between the config.yaml and snmp-base.yml

Obvious it's for y'all to make the call as to what you want design-wise, this is just something that could be a bit confusing and we'll need document the outcome on our side.

PR #451, opened 2022-10-13
Touch config file vs actually writing it out.

i3149:
This moves from actually writing the snmp config file when checking write permissions to just doing a touch on it. I think its a cleaner approach and fixes a bug where ktranslate crashing inside of a disco will corrupt the config file sometimes.


pkg/inputs/snmp/disco.go
```diff
@@ -320,6 +320,11 @@ func doubleCheckHost(result scan.Result, timeout time.Duration, ctl chan bool, m
 }
 
 func addDevices(ctx context.Context, foundDevices map[string]*kt.SnmpDeviceConfig, snmpFile string, conf *kt.SnmpConfig, isTest bool, log logger.ContextL, pollDuration time.Duration) (*SnmpDiscoDeviceStat, error) {
+	// If testing, just return if a touch works or not.
+	if isTest {
```
ehazlett, line 324: Where is `isTest` used?

PR #473, opened 2022-11-14
Allow tag values to be variables, filled in by snmp metadata

### LINKED ISSUE #469: FR: Add a mechanism for promoting attributes to user_tags
As a user, I need the ability to mark specific attributes for `user_tag` elements so they are shipped to NR with the `tags.*` prefix. This allows me to take advantage of common fields from SNMP in various ways in the NR UI, like faceting and filtering my devices.

A solution could look something like this:

```yaml
global:
  user_tags:
    location: !"SysLocation"
```

```yaml
devices:
  deviceOne_192.168.0.1:
    ...
    user_tags:
      location: !"SysLocation"
```

this would promote the value of the `SysLocation` field (`1.3.6.1.2.1.1.6`) to a field named `tags.location` before exporting the payload to the NR API.

i3149:
Given a tag selection like:

```
      user_tags:
        level: provider
        test: one
        type: nas
        location: "$SysLocation" 
```

This will replace `$SysLocation` in the tag value with whatever is found in snmp Metadata. For example, `"tags.location": "Unknown",`. 

Any tag value with a prefix of `$` will be interpreted as a variable. 
If the variable isn't found, this tag is removed. @thezackm  -- what do you think about this behavior? Or can just keep the `$whatever` value if not found. 

Closes #469

### DIFF
--- a/pkg/inputs/snmp/metadata/poll.go
+++ b/pkg/inputs/snmp/metadata/poll.go
@@ -259,10 +259,8 @@ func (p *Poller) toFlows(dd *kt.DeviceData) ([]*kt.JCHF, error) {
 	p.conf.SetUserTags(cs)
 	for k, v := range cs {
 		if strings.HasPrefix(v, TagValuePrefix) { // See if we can find this value in the MDS instead.
-			search := v[len(TagValuePrefix):]
-			v = "" // 0 tag out here incase of no match.
 			for tag, value := range dst.CustomStr {
-				if tag == search {
+				if tag == v[len(TagValuePrefix):] {
 					v = value
 				}
 			}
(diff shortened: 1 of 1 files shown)


Mesverrum:
I think dropping the tag for nulls is better than passing across the variable name, thats extra data on ingest that could be confusing.  Im imagining someone trying to `select uniques() ` and getting a mess.

RyanKadri (approved):
I'm still a newbie to both `ktranslate` and Go. But this seems like a good feature to me. I'd lean toward dropping tags when we can't find a value

thezackm:
how is a user to understand that their `SysLocation` field is null if we drop their entire `tags.location` and return nothing at all? 

IMO; respecting the data sovereignty means that we keep the placeholder for the tag and pass an empty string or null as appropriate. The user intends for there to be a tag, we cannot assume that they only want the tag IF there is an associated value. 

Adding that type of feature (drop if null) would be a solid compromise too.

i3149:
The challenge is that there's top level code which drops every attribute which is the empty string before it is sent out from New Relic. One idea is that we just keep the `$SysLocation` value if SysLocation isn't found. Then the user gets a signal that things are working, but also that there's no value for this in this instance. Also would fix a corner case where a user already has a tag set with a `$` prefix.

PR #475, opened 2022-11-15
Kentik Branch

i3149:
Make a branch where kentik specific stuff can be tested out.

### DIFF
--- a/.github/workflows/publish-kentik.yml
+++ b/.github/workflows/publish-kentik.yml
@@ -1,0 +1,74 @@
+name: Publish Kentik
+
+on:
+  push:
+    branches:
+      - kentik
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-kentik.yml
```diff
@@ -0,0 +1,74 @@
+name: Publish Kentik
```
tdanner, line 1: This file is almost an exact duplicate of `publish-staging.yml` (and quite similar to other `publish-*.yml` files). At some point it would be good to refactor and reduce duplication.

tdanner (approved):

.github/workflows/publish-kentik.yml
i3149: Hah! This is a very valid point. Probably can combine into some kinda action but man do I hate writing this stuff.

PR #476, opened 2022-11-15
feat: add minimalist config sample for meraki polling

thezackm:
adding a sample config file to show a minimalist config for polling the Meraki API in a New Relic setup.


i3149 (approved):
LGTM

PR #480, opened 2022-11-18
Add support for a ddog sink

i3149:
Allow ktranslate to export flow and snmp data to DataDog. Run by setting these flags: `-sink ddog -format ddog`. Also set this env variable: `DD_API_KEY`.

Sends metrics using the V2 API.


tdanner (approved):
I didn't test it and I'm not familiar with the datadog go sdk, but this looks reasonable to me.

i3149:
One more time plz.

tdanner (approved):

PR #483, opened 2022-11-22
Optionally monitor config file and exit if changed

i3149:
Monitor the config file optionally and shut down if it changes. 

Running this with the docker `restart=always` option will allow the system to pick up config changes automatically.


tdanner (approved):
Nice feature

PR #484, opened 2022-11-23
Force src addr for NR

i3149:
What do you think about this? Forces `src_addr` if present for all NR metrics. 

Separately, it will try to keep valid utf8 chars in traps and not strip out the whole string. The major problem here is that its probably being sent in some weird encoding scheme but I have no idea what it is. This is probably the best we can do for now.

### DIFF
--- a/pkg/formats/util/util_test.go
+++ b/pkg/formats/util/util_test.go
@@ -349,38 +349,53 @@ func TestCopyAttrforSNMP(t *testing.T) {
 	}
 	name := kt.MetricInfo{Oid: "oid", Mib: "mib"}
 
-	res := CopyAttrForSnmp(input, "test", name, nil, true)
+	res := CopyAttrForSnmp(input, "test", name, nil, true, true)
 	assert.Equal(len(input)+3, len(res)) // adds in three keys
 	assert.Equal("oid", res["objectIdentifier"])
 
 	for i := 0; i < MAX_ATTR_FOR_SNMP+10; i++ {
 		input[fmt.Sprintf("XXX%d", i)] = i
 	}
-	res = CopyAttrForSnmp(input, "test", name, nil, true)
+	res = CopyAttrForSnmp(input, "test", name, nil, true, true)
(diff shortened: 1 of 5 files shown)


Mesverrum (approved):

pkg/inputs/snmp/traps/traps.go
```diff
@@ -203,12 +203,13 @@ func (s *SnmpTrap) handle(packet *gosnmp.SnmpPacket, addr *net.UDPAddr) {
...
 				}
-				if utf8.ValidString(value) {
-					if res != nil {
-						dst.CustomStr[res.GetName()] = value
-					} else {
-						dst.CustomStr[v.Name] = value
-					}
+				if !utf8.ValidString(value) { // Print out value as a hex string.
+					value = fmt.Sprintf("%x", v.Value)
+				}
```
RyanKadri, line 208: Minor nit: Should this just be `value` rather than `v.Value`?

pkg/formats/util/util_test.go
```diff
@@ -349,38 +349,53 @@ func TestCopyAttrforSNMP(t *testing.T) {
...
 	assert.Equal(nil, res["foo"], res)
+
+	input = map[string]interface{}{"src_addr": "10.10.1.1"}
+	name = kt.MetricInfo{Oid: "oid", Mib: "mib", Table: "somethingElse"}
+	res = CopyAttrForSnmp(input, "test", name, nil, true, false)
+	assert.Equal("10.10.1.1", res["src_addr"], res)
+
+	input = map[string]interface{}{"src_addr": "10.10.1.1"}
+	name = kt.MetricInfo{Oid: "oid", Mib: "mib", Table: "somethingElse"}
+	res = CopyAttrForSnmp(input, "Uptime", name, nil, true, true)
```
RyanKadri, line 394: I'm sure I'm just missing how this works but I would've guessed this would drop the src_addr

RyanKadri (approved):
LGTM. I think I might be missing why two tests behave differently but I think that's me just not understanding the rules

PR #486, opened 2022-11-29
Setting service name as a tag for most inputs. 

i3149:
Also dropping pid from this.

Idea is to make things more like snmp.


RyanKadri (approved):
LGTM. Thanks!

PR #490, opened 2022-12-01
sysDescr remove from hardcoded base collection and use it from profil…

### LINKED ISSUE #461: remove hardcoded device SNMP metadata
In the file `pkg/inputs/snmp/metadata/device_metadata.go:`

There are some number of hardcoded variables

```
	SNMP_device_metadata_oids = func() *orderedmap.OrderedMap {
		m := orderedmap.NewOrderedMap()
		m.Set(".1.3.6.1.2.1.1.1.0", SNMP_sysDescr)
		m.Set(".1.3.6.1.2.1.1.2.0", SNMP_sysObjectID)
		m.Set(".1.3.6.1.2.1.1.4.0", SNMP_sysContact)
		m.Set(".1.3.6.1.2.1.1.5.0", SNMP_sysName)
		m.Set(".1.3.6.1.2.1.1.6.0", SNMP_sysLocation)
		m.Set(".1.3.6.1.2.1.1.7.0", SNMP_sysServices)
		m.Set(".1.3.6.1.6.3.10.2.1.1.0", SNMP_engineID)
		return m
	}()
```

Their usage in the `func NewDeviceMetadata` overrides any use of the in the snmp-profiles.

For this reason, I am not able to use sysDescr variable in the snmp-profiles, especially to parse it using regex and extract the valuable metadata information which is there.

AFAIK, these hardcodes are used in two places:
- func NewDeviceMetadata
- func GetBasicDeviceMetadata

According to the comment in the code:
```
// Super basic loop to get info for discovery.
func GetBasicDeviceMetadata(log logger.ContextL, server *gosnmp.GoSNMP) (*kt.DeviceMetricsMetadata, error) {
```

the second one is used for discovery and we can leave it there. 

Can we remove these hardcoded variables from hardcoded collection and use purely snmp-profiles to rely on them?

dpajin-kentik:
…e definitions. This partially fix #461 by avoiding "sysDescr" OID from the list of hardcoded OIDs used as a base list for polling. There is no need to hardcode it here, when it can be used from profile, like `system-mib.yml` which is anyway used for every device.


i3149:
Any thoughts on this @RyanKadri

i3149:
What's the reason behind taking just sysdesc out here? I guess I'm confused why you would just do this and not the rest too? For example, look at https://github.com/kentik/ktranslate/compare/kentik? 

Can you give a little more color as to what you're hopping to achieve here? Is is bad to poll sysdesc?

dpajin-kentik:
I need to perform "conversion" with "regex" on the sysDescr and extract the information which is inside the sysDescr. In order to do it, I need to be able to specify sysDescr in the profiles and have it "usable" from profiles. At the moment, any configuration of the sysDescr OID inside the profiles is ignored, since it is added in the "hardcoded" way.

I removed only sysDescr as I don't have any specific needs from other hardcoded values and for the moment I can used them the way they are exposed. But in general case, it would be preferable using profiles instead of hardcoded behavior. As I don't understand the code fully and I don't know where this is used or important on some other places or behaviors,I just wanted to minimize the potential impact on other functionalities relying on this code.

i3149:
https://github.com/kentik/ktranslate/compare/kentik 

Does this get you more what you are looking for?

i3149 (dismissed):
¯\_(ツ)_/¯

I don't _think_ this will break anything so I guess its fine if you want to.

i3149 (approved):
LGTM!

PR #492, opened 2022-12-01
Adding basic ability to run a script over metrics

i3149:
This hooks up metrics into the starlark system. Let me know what you think.


i3149:
Good enough for me! I don't think this will crash anything and we can iterate for features going forward.

dpajin-kentik:
great thanks, I didn't have a chance to test it, but I certainly need this, I will test it soon and its easier when I have a container build ready.

PR #495, opened 2022-12-03
Wip for what a managed config system might look like

i3149:
Here's a rough outline for what a managed config system might look like. 

https://github.com/kentik/ktranslate/tree/wip-managed-config/pkg/config/local 
Is an example of something which looks for a config change on a local file and restarts things if it sees this.

https://github.com/kentik/ktranslate/tree/wip-managed-config/pkg/config/nr
is a stub where you can pass data to NR. 

Thoughts here, especially on this interface?

```
type ConfigManager interface {
	Run(context.Context, func(*ktranslate.Config) error) // Run takes a context and a callback function to call whenever there is a new update to process
	DeviceDiscovery(kt.DeviceMap)                        // called whenever there is a new snmp device discovery to parse.
	Close()                                              // Called on shutdown of ktrans.
}
```

### DIFF
--- a/cmd/ktranslate/main.go
+++ b/cmd/ktranslate/main.go
@@ -707,6 +707,8 @@ func applyFlags(cfg *ktranslate.Config) error {
 				cfg.FlowInput.PrometheusListenAddr = val
 			case "nf.mapping":
 				cfg.FlowInput.MappingFile = val
+			case "config_provider":
+				cfg.CfgManager.ConfigImpl = val
 			// configs
 			case "config", "generate-config":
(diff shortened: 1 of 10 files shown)


RyanKadri (approved):
Nice. I think this all seems reasonable. I guess when we're more ready to use this, we'd add in the config flag to set the provider?

i3149:
I realized that there wasn't a flag to set the config provider option. Now this is in place.

RyanKadri (approved):
LGTM. Thanks for adding this in!

PR #497, opened 2022-12-06
Don't overwrite snmp-base.yaml file on package upgrade

RyanKadri:
When I went to update my package manager install of KTranslate, I noticed that the `snmp-base.yaml` file got overwritten. I think this flag should be true to prevent that behavior?

### DIFF
--- a/ktranslate-package.yml
+++ b/ktranslate-package.yml
@@ -42,6 +42,7 @@ files:
     file: config/snmp-base.yaml
     mode: "0644"
     user: ktranslate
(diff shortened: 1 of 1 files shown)


i3149 (approved):
Makes sense to me. TIL about the `keep` option here. This is only a problem if we want to change the structure of the `snmp-base.yaml` file in a breaking way which we try not to. 

Another approach would be to copy the file out to another file on install if that file doesn't exist but this seems cleaner.

PR #498, opened 2022-12-06
Remove namespace attributes for influx

### LINKED ISSUE #493: FR: configurable option to strip namespace from tag and metric names
To avoid the limitation where tag name must be the same for the same OID across all the profiles, I suggest using the "namespace" as a way to make tag names unique. 
The names can be in the following format `namespace:tag_name` or `namespace:metric_name`. 
For example: `bgp:neighbor-ip` or `ospf:neighbor-ip`

However, it would be good if we can strip the `namespace` before sending further toward the outputs and having that as a configurable option.

Can we implement this for InfluxDB output?

i3149:
Trims the namespace token off tags for influx. Also a test for this. 

Closes #493

### DIFF
--- a/cmd/ktranslate/main.go
+++ b/cmd/ktranslate/main.go
@@ -425,6 +425,8 @@ func applyFlags(cfg *ktranslate.Config) error {
 			// pkg/formats/influxdb
 			case "influxdb_measurement_prefix":
 				cfg.InfluxDBFormat.MeasurementPrefix = val
+			case "influxdb_namespace_token":
+				cfg.InfluxDBFormat.NamespaceToken = val
 			// pkg/sinks/prom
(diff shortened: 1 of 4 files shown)


tdanner (dismissed):

dpajin-kentik:
@i3149 can we apply this to metrics as well, not just tags? Also, I am using ":" as a separator, not "::"

i3149:
You're not worried that `:` will be found inside metric or attribute? I went with `::` to (~moreso) limit this worry but can do `:` I guess. Also I started out with c++ and `::` forever.

i3149:
Easy enough to make this a config value with a default. Also will support metrics.

dpajin-kentik (approved):

PR #501, opened 2022-12-07
fix influxdb measurement name issue

dpajin-kentik:
my previous PR for influxdb measurement name behavior had a bug. This is a fix for that


i3149:
Let me land #498 and then we can write a test for this too

PR #504, opened 2022-12-09
Adding a way to check v1,2c, and 3 snmp

### LINKED ISSUE #494: FR: Allow discovery to iterate testing v3>v2>v1 automatically
As a user with a mixed environment where a small number of devices scattered throughout the subnets only support v1 it is a chore to separate out the v1 devices for their own discovery jobs.

From a customer "What I’d expect to see is: check snmpv3 > check snmpv2c >check snmpv1. Sadly, we do have SNMPv1 devices in our environment."

i3149:
Adds this option `doublecheck_snmp_v1: true` to the `discovery` section of the snmp config file. 

When set to true, v3, then v2c and finally v1 configs are checked for all devices. If false, current behavior remains. 

Closes #494

### DIFF
--- a/pkg/inputs/snmp/disco.go
+++ b/pkg/inputs/snmp/disco.go
@@ -241,27 +241,34 @@ func doubleCheckHost(result scan.Result, timeout time.Duration, ctl chan bool, m
 
 	// Loop over all possibe v2c options here if any are set.
 	if md == nil || md.SysObjectID == "" { // Only check these if v3 hasn't found anything.
-		for _, community := range conf.Disco.DefaultCommunities {
-			device = kt.SnmpDeviceConfig{
-				DeviceName: result.Name,
-				DeviceIP:   result.Host.String(),
-				Community:  community,
-				UseV1:      conf.Disco.UseV1,
-				Debug:      conf.Disco.Debug,
-				Port:       uint16(conf.Disco.Ports[0]),
-				Checked:    time.Now(),
+		versions := []bool{conf.Disco.UseV1} // By default, just check v1 or v2c only.
+		if conf.Disco.DoubleCheckV1 {
+			versions = []bool{false, true} // But, if we ask for it, first check v2c and then check v1.
+		}
+	outer:
+		for _, usev1 := range versions {
+			for _, community := range conf.Disco.DefaultCommunities {
+				device = kt.SnmpDeviceConfig{
+					DeviceName: result.Name,
+					DeviceIP:   result.Host.String(),
+					Community:  community,
+					UseV1:      usev1,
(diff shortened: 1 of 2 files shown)


pkg/kt/snmp.go
```diff
@@ -243,6 +243,7 @@ type SnmpDiscoConfig struct {
 	ReplaceDevices     bool            `yaml:"replace_devices"`
 	NoDedup            bool            `yaml:"no_dedup_engine_id,omitempty"`
 	CheckAll           bool            `yaml:"check_all_ips,omitempty"`
+	DoubleCheckV1      bool            `yaml:"doublecheck_snmp_v1,omitempty"`
```
RyanKadri, line 246: I think the specific flag name is tripping me up a bit. I feel like `doublecheck` might imply that something is literally being checked twice. Not sure about the best replacement name though. Maybe something like `check_all_snmp_versions`? That could be confusing if the user doesn't provide V3 info though. Maybe `use_snmp_v1_fallback`?

pkg/inputs/snmp/disco.go
```diff
@@ -241,27 +241,34 @@ func doubleCheckHost(result scan.Result, timeout time.Duration, ctl chan bool, m
...
-				log.Warnf("There was an error when starting SNMP interface component -- %v.", err)
-				return
-			}
-			md, err = metadata.GetBasicDeviceMetadata(log, serv)
-			if err != nil {
-				log.Warnf("Cannot get device metadata on %s: %v. Check for correct snmp credentials.", result.Host.String(), err)
-				continue
+		versions := []bool{conf.Disco.UseV1} // By default, just check v1 or v2c only.
+		if conf.Disco.DoubleCheckV1 {
+			versions = []bool{false, true} // But, if we ask for it, first check v2c and then check v1.
```
RyanKadri, line 246: I assume the answer is no for now. But do we care if the user has `UseV1 == false` and `DoubleCheckV1 == true`?

RyanKadri (commented):
Quick question on the flag name. Not sure about the best alternative though

pkg/kt/snmp.go
i3149: I like `check_all_snmp_versions`.

pkg/inputs/snmp/disco.go
i3149: I would go with no here for now. This is kinda bad naming with `UseV1` where it really should be 2 variables, `UseV1` and `UseV2`. But we are kinda stuck with just one. Idea being `DoubleCheckV1` (renamed to `CheckAllVersions`) means check everything, while use v1 just forces only a check of v1.

RyanKadri (approved):
LGTM

PR #515, opened 2022-12-15
Allow rates for influx

### LINKED ISSUE #472: FR: conversion rate and diff
As a user, I want to easily get the rate from the counter metrics and save them into separate metrics:

I propose that we use “conversion” methods for this, for example:
```yaml

metrics:
  - MIB: IF-MIB
    table:
      OID: 1.3.6.1.2.1.2.2
      name: ifTable
    symbols:
      - name: ifHCInOctets
        OID: 1.3.6.1.2.1.31.1.1.1.6
        tag: in-octets
        conversion:
          rate: "in-bps"
        or 
        conversion: "rate:in-bps"

      - name: ifInErrors
        OID: 1.3.6.1.2.1.2.2.1.14
        tag: in-errors
        conversion:
          diff: "in-errors-diff"
```

i3149:
What do you guys think of this -- uses the existing format field in profiles to support a `rate` format. Then at format time it gets calculated out. 

Closes #472

### DIFF
--- a/pkg/formats/influx/influx.go
+++ b/pkg/formats/influx/influx.go
@@ -545,7 +545,7 @@ func (f *InfluxFormat) fromSnmpDeviceMetric(in *kt.JCHF) []InfluxData {
 					Tags:        attrNew,
 				})
 			case kt.Rate:
-				delta, found := f.getDelta(m, attrNew, in.CustomBigInt[m], in.CustomBigInt["Uptime"])
+				delta, found := f.getDelta(m, attrNew, in.CustomBigInt[m], in.CustomBigInt["uptime_delta"])
 				if found {
 					results = append(results, InfluxData{
 						Name:      f.config.MeasurementPrefix + mib,
--- a/pkg/formats/influx/influx.go
+++ b/pkg/formats/influx/influx.go
@@ -545,7 +545,7 @@ func (f *InfluxFormat) fromSnmpDeviceMetric(in *kt.JCHF) []InfluxData {
 					Tags:        attrNew,
 				})
 			case kt.Rate:
-				delta, found := f.getDelta(m, attrNew, in.CustomBigInt[m])
+				delta, found := f.getDelta(m, attrNew, in.CustomBigInt[m], in.CustomBigInt["Uptime"])
 				if found {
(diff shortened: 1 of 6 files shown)


i3149:
For example:

```
      # Synology disk temperature (C)
      - name: diskTemperature
        OID: 1.3.6.1.4.1.6574.2.1.1.6
        format: rate
```

pkg/formats/influx/influx.go
```diff
@@ -534,14 +536,25 @@ func (f *InfluxFormat) fromSnmpDeviceMetric(in *kt.JCHF) []InfluxData {
...
 					Timestamp:   in.Timestamp,
 					Tags:        attrNew,
 				})
-			} else {
+			case kt.Rate:
+				delta, found := f.getDelta(m, attrNew, in.CustomBigInt[m])
+				if found {
+					results = append(results, InfluxData{
+						Name:      f.config.MeasurementPrefix + mib,
+						Fields:    map[string]int64{metricName: delta},
```
tdanner, line 552: Don't you need to divide the delta by the time since the previous poll to convert a counter to a rate? Reporting the counter delta but not the time delta will just force the backend to figure out the time delta. Easier to do it here where that state can be kept in memory.

tdanner:
diskTemperature is almost certainly a gauge, not a counter or rate!

In the context of the SNMP profile yaml, I think it would make more sense to label a counter as a counter. What do you think?

pkg/formats/influx/influx.go
i3149: That's a good q!  The doing it at query time way is one approach. For example, NR has a few neat functions like: [func-rate](https://docs.newrelic.com/docs/query-your-data/nrql-new-relic-query-language/get-started/nrql-syntax-clauses-functions/#func-rate) which lets you mess with things to see both rate and not rate without storing data twice. 

Since there's a uptime ticker here anyway I can just add it in.

i3149:
> diskTemperature is almost certainly a gauge, not a counter or rate!

You make a good point -- this is just me showing how I'm testing rate functionality, not that it should itself be a rate. Really though, anything can be a rate if you try hard enough :)

Issue #509, opened 2022-12-13
snmp-profiles: tags enum values in upper case, sometimes returned in lowercase


dpajin-kentik:
I have an issue that the metric tag as enum with capital letters is sometimes returned as a lowercase 
```

      - tag: components:type
        column:
          OID: .1.3.6.1.2.1.47.1.1.1.1.5
          name: entPhysicalClass
          enum:
            OTHER: 1
            UNKNOWN: 2
            CHASSIS: 3
            BACKPLANE: 4
            CONTAINER: 5
            POWER_SUPPLY: 6
            FAN: 7
            SENSOR: 8
            MODULE: 9
            PORT: 10
            STACK: 11
            CPU: 12
```
The values returned are sometimes in lowercase:
I don't have any other profile with the same OID or same 
```
/components,description=Transceiver\ Port\ Container\ Gi1/1,device_ip=198.186.190.9,device_name=cat6509,index=1025,name=Transceiver\ Port\ Container\ Gi1/1,parent=1024,parent-position=1,removable=false,type=container fru=2i 1670939767
/components,description=module\ 2\ power-output-fail\ Sensor,device_ip=198.186.190.9,device_name=cat6509,index=2002,name=module\ 2\ power-output-fail\ Sensor,parent=2000,parent-position=1,removable=false,type=sensor fru=2i 1670939767
/components,description=Te2/3,device_ip=198.186.190.9,device_name=cat6509,index=2034,name=Transceiver\ Port\ Te2/3,parent=2000,parent-position=3,removable=false,type=port fru=2i 1670939767
/components,description=WS-F6K-PFC3BXL\ Policy\ Feature\ Card\ 3\ sub-module\ of\ 1,device_ip=198.186.190.9,device_name=cat6509,index=1021,mfg-name=Cisco\ Systems,model=WS-F6K-PFC3BXL,name=WS-F6K-PFC3BXL\ Policy\ Feature\ Card\ 3\ Rev.\ 1.4,parent=1020,parent-position=1,removable=true,serial-no=SAL09169NV9,tags.type=spine,type=MODULE fru=1i 1670939767
[...]

PR #523, opened 2023-02-01
Fix nondeterministic enum indexing (#509)

kentik-will:
The MIB profile loading code modifies the content of the Enum field while iterating over it. This is legal, but the results are [nondeterministic](https://github.com/golang/go/issues/9926), added items may or may not be included in the iteration. This is the cause of #509, which occasionally uses the newly-added lowercase value as the reverse index, instead of the original uppercase value.

### DIFF
--- a/pkg/inputs/snmp/mibs/profile.go
+++ b/pkg/inputs/snmp/mibs/profile.go
@@ -447,13 +447,11 @@ func (p *Profile) GetMetrics(enabledMibs []string, counterTimeSec int) (map[stri
 				AllowDup:     metric.Symbol.AllowDup,
 				Condition:    metric.Symbol.GetCondition(p),
 			}
+
 			if len(mib.Enum) > 0 {
-				mib.EnumRev = make(map[int64]string)
+				indexEnum(mib)
 			}
-			for k, v := range mib.Enum {
-				mib.Enum[strings.ToLower(k)] = v
-				mib.EnumRev[v] = k
-			}
+
 			if metric.Symbol.PollTime > 0 {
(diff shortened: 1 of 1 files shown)


tdanner (approved):

dpajin-kentik:
@kentik-will, @tdanner  is this fix merged? I see that the issue is closed, but I used the latest version of the ktranslate and I see that the issue behavior is still there, so I would imagine that this was never merged?

tdanner:
No, this hasn't been merged. I'll merge it now.

kentik-will:
@tdanner too late for this one, but for future PR merges what do you guys think about having a branch other than `main` where we can test on prior to merging to `main`? This change had only brief testing and I'm still fairly unfamiliar with the code base, so merging directly in to main is a bit risky.

tdanner:
A branch for testing/staging sounds reasonable to me. What should we call it? I see in https://github.com/kentik/ktranslate/branches that Ian at one point had a `develop` branch, but it hasn't been touched in a couple of years.

kentik-will:
How about naming it `kentik` and we can set up the CI/CD to push docker images to the `kentik` tag?

tdanner:
As a matter of fact, we already have a `kentik` branch with CI/CD set to push docker images to the `kentik` tag. Using that to try out changes sounds fine to me. Note that the `kentik` branch and docker image tag bake in https://github.com/kentik/collection-profiles instead of https://github.com/kentik/snmp-profiles.

kentik-will:
Great, let's use the `kentik` branch then. I'll continue to make PRs against `main`, but we can pull them into the `kentik` branch for testing.

Issue #489, opened 2022-12-01
FR: PollingHealth and ping metric measurement name


dpajin-kentik:
Currently PollingHealth metric is submitted as “computed” measurement name. 
I am not sure how intuitive this is, would it be possible to change it to “/device/snmp-polling”?

Same for ping results - currently they appear separated, RTTs metrics are separated from PacketLoss metrics. Maybe we can glue them together into /device/ping?

dpajin-kentik:
if I may add one more thing related to the computed measurement and PollingtHealth. PollingHealth metric is not very convenient if it is "alone" as it has multiple states and does not give a convenient metric to, for example, calculate the availability of the device. 

SInce pollinghealth also encodes errors, I would like to have an additional metric which would just provide two stated, good and bad. 
For example, metric "snmp-status" which would be:
- device polling works: value 1.
- device polling does not work: value 0.

This gives me an easy alerting approach and easy way to calculate availability over time, as I can only look for average value of this metric over time to get the availability. I would suggest doing the similar approach for Ping testing.

dpajin-kentik:
@kentik-will is this something that you can help with?
cc: @tdanner

tdanner:
@dpajin-kentik - could you edit the description of this issue to list the expected measurements and metrics? I think that would be easier to work from. Or create a dedicated page for that and link it here. As it is, this is phrased in terms of a set of changes and it is not clear what the full picture should look like.

kentik-will:
The changes in PR #532 will allow you to add profile entries to override the `computed` name, e.g.:

```yaml
metrics:
  - MIB: KENTIK-KTRANSLATE
    table:
      OID: .1.2.3.1
      name: /device
  - MIB: KENTIK-KTRANSLATE
    table:
      OID: .1.2.3.2
      name: /device/snmp-polling
  - MIB: KENTIK-KTRANSLATE
    table:
      OID: .1.2.3.3
      name: /device/ping
```

Does this work for you?

dpajin-kentik:
@kentik-will, I just tested this and it works, thanks.
However, I have one comment before closing this issue. These special OIDs that we use here, I would like to avoid use any random values which can overlap with something and go with something more "defined". Kentik has registered its enterprise number assignment with IANA (https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers) and its number is 57558, so I think it would be more appropriate to use OIDs from that part of the MIB tree, even for this purpose.

[...]

PR #525, opened 2023-02-14
Merge ping metrics into a single record

kentik-will:
Retain src_addr field for all ping metrics so the merging process will include all metrics in a single output record. 

This is one of a number of different features requested in #489.


tdanner (approved):
Looks fine. Have not tested.

PR #527, opened 2023-02-16
Change profile source for kentik variant

tdanner:
The kentik variant should use the collection-profiles repo instead of the kentik branch of the snmp-profiles repo.


kentik-will (approved):
LGTM

PR #532, opened 2023-02-28
Make computed device metric names configurable

kentik-will:
Add dummy OIDs for the synthetic metrics ktranslate generates for uptime, health status, and ping, then output the table name associated with those OIDs, if one exists in the profile, instead of `computed`.

This is one of the features requested in #489.


kentik-will:
Also added the requested `PollingStatus` metric from #489 since it depends on this PR.

i3149 (approved):
LGTM

PR #544, opened 2023-03-16
Adding assume role for S3 sink

Let-itGo:
The change drives from the requirement of having IAM role ARN which has write permissions to the S3 bucket.


Let-itGo:
Hi @i3149 , Can you have an initial review about this?
can this change be added to this repo?

PS: This is my first commit to go and docker.

pkg/sinks/s3/s3.go
```diff
@@ -75,8 +80,38 @@ func (s *S3Sink) Init(ctx context.Context, format formats.Format, compression kt
...
+			return fmt.Errorf("Session is not created with region %q", s.config.Region)
+		}
+		// sess := session.Must(session.NewSession(&aws.config{Region: aws.String(s3Region), Credentials: creds}))
+		s.client = s3manager.NewUploader(sess)
+	} else {
+		sess := session.Must(session.NewSession())
+		s.client = s3manager.NewUploader(sess)
+	}
+
+	// s.client = s3manager.NewUploader(sess)
```
i3149, line 114: Take this comment out, along with the other commented line.

i3149 (changes requested):
Looks good for your first golang pr! I was thinking you can use env variables for the S3 Role but looking through the SDK docs it doesn't look like this would work. Your approach seems good to me. Can you clean up the few commented lines and then I'll merge in.

pkg/sinks/s3/s3.go
Let-itGo: Sure, working on it

Let-itGo:
Hi @i3149 , I have added further details here. I am noticing that s.send (line 121) is not being executed in goroutine and same is "go s.tmp_credentials()" (line 90). 
Can you help me with the modification where we can call this function before tmp credentials are expired or maybe after every 15 mins?

i3149:
> Hi @i3149 , I have added further details here. Can you check the changes.
> 
> Also, I'm not able to see the data in the "s.buf.Len()" of S3 Sink, Can you help me with that?

You should be able to do something like `s.buf.Bytes()` to get the data in this buffer. You can't directly read the data, but this will return a pointer to the backing slice. https://pkg.go.dev/bytes#Buffer.Bytes

i3149 (dismissed):
What do you think about this refactor: 

https://github.com/kentik/ktranslate/blob/pye-s3/pkg/sinks/s3/s3.go

This will raise an error if the first time through the AWS creds do not work and also fixes a race condition where the client might not be initialized before its used the first time.

Let-itGo:
Woah! that's a nice update. Thanks ✌🏼

Let-itGo:
Hi @i3149 , What is the merge process now?

i3149:
> Hi @i3149 , What is the merge process now?

Its cleanest if you just take that file and commit it on your branch. Then I can approve the whole PR as a unit, keeping you the author.

Let-itGo:
Hi @i3149 , I have committed the changes as per your branch:
https://raw.githubusercontent.com/kentik/ktranslate/pye-s3/pkg/sinks/s3/s3.go

i3149 (approved):
Nice work!

PR #548, opened 2023-03-24
Fixing panic when variable tags are the empty string

i3149:
There's a bug in how this variable mapping is applied. There's this loop here: 

```
            for tag, value := range dst.CustomStr {
                if tag == v[len(TagValuePrefix):] {
                    v = value
                }
            }
```

 the first time through, it will swap v to the value and everything is good. The problem is that the loop isn't exiting and v now is the empty string.

next time though, it will try to take a slice of the empty string and get your stack trace.
Code fix is to add a break statement so that once a correct variable is applied, no more variables will be swapped in.
for now, I'd recommend if possible you remove any of the $SysContact in tags for the offending containers
this should prevent the crash


kentik-will (approved):
LGTM

PR #552, opened 2023-03-28
Report interface counters in addition to deltas

tdanner:
Consider this an experiment for now. What this does: for each of the interface counters (ifHC{In,Out}BroadcastPkts, ifHC{In,Out}MulticastPkts, ifHC{In,Out}Octets, ifHC{In,Out}UcastPkts, if{In,Out}Discards, if{In,Out}Errors) that ktranslate normally returns in delta form, this also returns the raw counter value with the same metric name and a `_counter` suffix.


i3149 (approved):
Can't hurt.

tdanner:
Let's go ahead and merge this into the kentik branch. I don't know if this should ever go to `main`, but I'd like to experiment with it wider than just my copy.

PR #553, opened 2023-03-29
improve netflow format export

pyke369:
- auto-limit flows batch size to avoid reaching network message size limit
- pack CompanyId and DeviceId fields into observationPointId netflow entity


i3149 (approved):
Looks good to me. My only question is do we want to also have a loop in the netflow format which will ensure that the packet size never goes over a fixed size?

pyke369:
I dont think it necessary anymore, no.

i3149:
Lets merge this in then :)

PR #559, opened 2023-04-20
Allow named match metrics

i3149:
A request came in to support named matches as metrics in addition to metadata. Here it is. Test has example input, which will get turned into something like:

```
 "name": "kentik.snmp.Exhaust",
        "type": "gauge",
        "value": 31,
```


Mesverrum (approved):

Mesverrum:
For the sake of documenting the syntax, this is using regex capture groups and the profile would be something like?
symbol: 
  name: abc
  oid: 1.2.3.
  conversion: "regexp::.*Intake.*?(?P<Intake>\d+) .*Exhaust.*?(?P<Exhaust>\d+)"

i3149:
It would be like this:

```
symbol: 
  name: abc
  oid: 1.2.3.
  conversion: "regexp:.*Intake.*?(?P<Intake>\\d+) .*Exhaust.*?(?P<Exhaust>\\d+)"
```

Just like yours but there's a double `\\` for all of backslashes.

PR #561, opened 2023-04-26
Drop ping intervals where the number of recieved packets is too high

### LINKED ISSUE #555: Bug: inconsistent kentik.ping.PacketLossPct values when customer has two containers for same device
**Supplemental info provided in backchannel.**

Customer’s config `kentik.ping.PacketsSent` is consistently 6 (I’m thinking they do 10 second pings) and then when the numbers get wonky we have a case where one data point shows  5 packets received and the next interval sent 6 but received 7.

When you have more received than sent ktrans is returning
``5 of 6
{“type”:“gauge”,“count”:1,“sum”:16.666,“min”:16.666,“max”:16.666,“latest”:16.666}
7 of 6
{“type”:“gauge”,“count”:1,“sum”:-9.223372036854776E15,“min”:-9.223372036854776E15,“max”:-9.223372036854776E15,“latest”:-9.223372036854776E15}``

Responses seem to be falling into the wrong buckets and its making the math incorrect.

**Supplemental info provided in backchannel.**

i3149:
Adds a guard and if the number of received packets is too high for an interval skip this interval when reporting ping values. 

Closes #555

### DIFF
--- a/pkg/inputs/snmp/metrics/device_metrics.go
+++ b/pkg/inputs/snmp/metrics/device_metrics.go
@@ -405,7 +405,7 @@ func (dm *DeviceMetrics) GetPingStats(ctx context.Context, pinger *ping.Pinger) 
 	dm.ping.sent = sent
 	dm.ping.received = received
 	percnt := 0.0
-	if diffSent > 0 {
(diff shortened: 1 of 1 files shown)


Mesverrum (approved):

i3149:
I'm not sure the best approach here. Dropping problematic cycles is what this does, just because that was easiest to implement. Open to other ideas however.

PR #563, opened 2023-04-28
Using snmp engine id as alternative to ip for traps

### LINKED ISSUE #558: FR: Allow SNMPv3 devices to use EngineID instead of IP as their identifying attribute
Ran into a situation with a customer today where they are sending SNMPv2 traps through a load balancer into a k8s hosted container.  The ktranslate container is seeing the src_addr as the internal k8s ip of the load balancer instead of seeing the true source IP and we don't appear to have the option to preserve the source IP for this scenario.

After some poking around we found that with SNMPv3 traps there *should* be an EngineID that we can use to identify the source, but we would also need ktranslate to gather the be able to attempt to match incoming traps against the stored engineid in addition to the way it currently does ip to device_name mapping.

i3149:
Lets SNMP v3 Context Engine ID be used to look up devices in addition to the sending IP when processing a trap. 

Closes #558

### DIFF
--- a/pkg/inputs/snmp/traps/traps.go
+++ b/pkg/inputs/snmp/traps/traps.go
@@ -148,12 +148,15 @@ func (s *SnmpTrap) Listen() {
 }
 
 func (s *SnmpTrap) handle(packet *gosnmp.SnmpPacket, addr *net.UDPAddr) {
-	s.log.Infof("got trapdata from %s", addr.IP)
+	s.log.Infof("got trapdata from %s, EngineID %s", addr.IP, packet.ContextEngineID)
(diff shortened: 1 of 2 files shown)


i3149:
Tested with 

```
 snmptrap -e 0102030405 -v 3 -E 0xffffffffff -u myuser -a MD5 -A mypassword -l authNoPriv 127.0.0.1:1620 '' 1.3.6.1.4.1.8072.2.3.0.1 1.3.6.1.4.1.9.9.187.1.3.4.5.22.33 i 123456
```

And a device with `engine_id: ff:ff:ff:ff:ff`

Mesverrum (approved):

PR #565, opened 2023-05-02
Tweak filters

i3149:
Reworking filters to make them more logical. Now you can just say `-filters fooII,==,12` and things will work to filter out all events where fooII doesn't equal 12.

### DIFF
--- a/cmd/ktranslate/main.go
+++ b/cmd/ktranslate/main.go
@@ -401,7 +401,11 @@ func applyFlags(cfg *ktranslate.Config) error {
 				}
 				cfg.EnableHTTPInput = v
 			case "enricher":
-				cfg.EnricherURL = val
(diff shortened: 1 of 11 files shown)


i3149:
Now you can do things like `-filter="foo,==,no or fooII,==,12"`. Other wise, `-filter` flags are ANDed together. You can also set a flag `-enricher=/path/to/script.py` to filter this way, setting anything you do not like to None. For example:

```
def main(n):
    keep = 0
    for i in range(len(n)):
      if n[i].TrapOID == ".1.3.6.1.4.1.8072.2.3.0.1":
          print(n[i])
          keep = keep+1
      else:
         n[i] = None

    return keep
```

Mesverrum (approved):

PR #571, opened 2023-05-04
Bringing the release version up to date

i3149:
I think this works to accurately get tags into the version of ktranslate which we make when a release happens.


RyanKadri (approved):
LGTM

PR #576, opened 2023-05-12
Allowing an override to let non increasing oids be returned.

### LINKED ISSUE #554: Attempting to poll certain OIDs return log message `<OID> failed to return results` - tcpdump to PCAP shows "Fragmented IP protocol"
The profile `elemental/elemental-device.yml` in the `kentik/snmp-profiles` GitHub repository has the following OIDs:
```
  - MIB: ELEMENTAL-LIVE-MIB
    table:
      OID: 1.3.6.1.4.1.37086.5.1.1
      name: eventTable
    symbols:
      - name: eventRunning
        OID: 1.3.6.1.4.1.37086.5.1.1.1.3
        enum:
          notRunning: 0
          running: 1
      - name: eventError
        OID: 1.3.6.1.4.1.37086.5.1.1.1.4
        enum:
          noError: 0
          error: 1
    metric_tags:
      - column:
          OID: 1.3.6.1.4.1.37086.5.1.1.1.2
          name: eventName
      # The numerical ID of the node that the event is running on.
      - column:
          OID: 1.3.6.1.4.1.37086.5.1.1.1.6
          name: nodeId
```
([source](https://github.com/kentik/snmp-profiles/blob/main/profiles/kentik_snmp/elemental/elemental-device.yml#L48-L70))

When Ktranslate attempts to poll those OIDs, the following message is shown in the container's logs:
```
{"log":"2023-03-11T16:00:06.299 ktranslate/neteng_elemental2 [Warn] KTranslate\u003elonelemental02 OID 1.3.6.1.4.1.37086.5.1.1.1.6 failed to return results, Metric Name: nodeId\n","stream":"stdout","time":"2023-03-11T16:00:06.299316892Z"}
```

The OID and Metric Name change depending on the OID being used, but data is never returned from these. In the process of debugging the issue, we grabbed a PCAP file using tcpdump and found what we assume is the source of the issue. Any time an OID from that profile list is polled, the returned packets are fragmented:

```
Fragmented IP protocol (proto=UDP 17, off=0, ID=8c63)
```

The ID changes depending on the packet, but the message stays the same. PCAP can be provided if needed, but will need to be done so privately in case it contains private information. Polling the device directly works and the data is returned correctly. Attempting to switch the SNMP version being used via the config file from v2c to v1 doesn't change anything. 

[fragmentedPacketResources.zip](https://github.com/kentik/ktranslate/files/11163470/fragmentedPacketResources.zip)

i3149:
Need to test more but this could be a fix for #554 . 

Closes #554 .

### DIFF
--- a/pkg/inputs/snmp/metadata/device_metadata.go
+++ b/pkg/inputs/snmp/metadata/device_metadata.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net"
 	"sort"
+	"strings"
 	"unicode/utf8"
 
 	"github.com/elliotchance/orderedmap"
@@ -105,7 +106,12 @@ func (dm *DeviceMetadata) poll(ctx context.Context, server *gosnmp.GoSNMP) (*kt.
 		if !mib.IsPollReady() { // Skip this mib because its time to poll hasn't elapsed yet.
 			continue
 		}
-		oidResults, err := snmp_util.WalkOID(ctx, dm.conf, oid, server, dm.log, "CustomDeviceMetadata")
+		walkOid := oid
+		if mib != nil && mib.WalkTable {
+			walkOid = mib.TableOid
+			dm.log.Debugf("Walking %s as a table %s", mib.Name, walkOid)
+		}
+		oidResults, err := snmp_util.WalkOID(ctx, dm.conf, walkOid, server, dm.log, "CustomDeviceMetadata")
 		if err != nil {
 			dm.metrics.Errors.Mark(1)
 			continue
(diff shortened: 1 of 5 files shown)


i3149:
This has two new options to try and make the AWS elemental mib work. 

1) `no_check_increasing_oids: true`. Set this in the device config per device.
2) `walk_full_table: true`. Set this per mib table in a snmp profile. 

Its hard to test without an actual device but my best guess is that 2 is all that is needed. I'll set up a seperate PR with this.

i3149:
Connected with https://github.com/kentik/snmp-profiles/pull/664

PR #583, opened 2023-05-24
Updating Assume Role in S3 (AWS)

Let-itGo:
Updating the instance profile role reading as we don't have to explicitly update the temporary credentials of the Instance profile. It's handled by the sts internally.


Let-itGo:
Hi @i3149 , Can you please review this change?
It's updating PR: https://github.com/kentik/ktranslate/pull/544/ .

i3149 (approved):
LGTM

PR #586, opened 2023-06-13
fix(MM-Script): Check MaxMind key and init config dir

indigo423:
Check if the user has set the MaxMind download key and if not give some hints on how to set it and where he can sign up to get one. Initialize the config directory so the user doesn't have to create it himself.


i3149 (approved):
Nice contribution! thanks.

PR #587, opened 2023-06-16
Various hacks to make ktranslate not waste time on down devices

tdanner:
This is going to the `kentik` branch of ktranslate, not `main`.

What I was seeing is that if ktranslate is happily polling SNMP device and it goes down, there was extreme latency (15 minutes or more) before it would be reported down because ktranslate was waiting for each... OID... request... to... timeout... including... retries...

So this PR uses a `context.WithTimeout` around the whole metrics polling process. I also changed the status values for several conditions to all be `SNMP_BAD_POLL_TIMEOUT`. I'm not confident that this is the right choice overall, but it works for my needs and will hold us over until we can switch to ranger.

### DIFF
--- a/pkg/inputs/snmp/metrics/poll.go
+++ b/pkg/inputs/snmp/metrics/poll.go
@@ -211,7 +211,9 @@ func (p *Poller) StartLoop(ctx context.Context) {
 					continue
 				}
 
-				flows, err := p.Poll(ctx)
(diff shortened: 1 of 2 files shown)


i3149 (approved):
Smart! Good idea to add this in. 

A further refinement could be keeping an internal list of up/down devices based on the ping responses ktrans is getting in another thread, but this is probably too complicated what we need here now.

PR #588, opened 2023-06-16
Merge Kentik Branch Into Main

i3149:
Look at merging the Kentik branch into the main section.


tdanner (approved):
This looks fine. Probably better to not just have these changes hanging out there indefinitely. However do note that I expect recon to switch from ktranslate to ranger in a month or two.

i3149:
Cool, this is kinda just tieing all the loose ends together. I think ktranslate is going to halt development going forward so will be a good place to leave things.

PR #589, opened 2023-06-23
Reverse DNS fallback

RyanKadri:
When determining the `device_name` for network syslogs, if there is a DNS resolver to look up hostnames from the sender's IP, but the resolver does not give a response, we should fallback to the IP address


i3149 (approved):
Great idea!

PR #601, opened 2023-08-17
Upgrading some dependencies to fix security scans

i3149:
Upgrading x/net and prometheus packages to try and resolve a few medium security advisories.


thezackm (approved):
🔐 🚀

PR #604, opened 2023-08-24
add the device status call for meraki

### LINKED ISSUE #603: Feature request for Meraki API queries
Would love to be able to optionally also get Meraki device status from a query like
https://developer.cisco.com/meraki/api-v1/get-organization-devices-statuses/
and returning at least the name, serial, status, productType, and model
where the query can be limited by the network(s) specified in the yaml file
and the product type(s) can also be specified there.
Because of the scale of my organization, I am NOT looking for any usage or client data along with this.
Thank you so much for your consideration.

i3149:
This adds in https://developer.cisco.com/meraki/api-v1/get-organization-devices-statuses/ to the mix of API calls the meraki adapter understands. To turn on, set 

`monitor_devices: true` AND 
```
preferences:
           device_status_only: true
```

OR just:

```
preferences:
           device_status: true
```

Data looks like

```
      {
        "name": "kentik.meraki.device_status.Status",
        "type": "gauge",
        "value": 1,
        "attributes": {
          "mib-name": "meraki",
          "serial": "Serial",
          "network_id": "L_NETWORK",
          "org_id": "ID",
          "instrumentation.name": "meraki.device_status",
          "mac": "dead::beef",
          "device_name": "name",
          "provider": "meraki-cloud-controller",
          "eventType": "KSnmpDeviceMetric",
          "objectIdentifier": "meraki",
          "network": "Network",
          "last_reported_at": "2023-08-24T03:21:23.327000Z",
          "product_type": "wireless",
          "org_name": "MyOrg",
          "src_addr": "1.1.1.1",
          "status": "online",
          "model": "MR46E",
          "tags": "TAG1,TAG2"
        }
      },
```

The metric has a value of 1 if status is online, 0 otherwise. 

QQ -- the `device_status` is getting complicated. Open to other ideas here on how to manage. Idea being that there are device status (the APs/switches) and the device's clients status (things connected to them). 

Filter by these options in the config file:

```
          networks:
           - "My Site 21"
         product_types:
           - wireless
``` 

Closes #603

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/json-iterator/go v1.1.12
 	github.com/judwhite/go-svc v1.2.1
 	github.com/kentik/api-schema-public v0.0.0-20220322181339-896729e59945
-	github.com/kentik/dashboard-api-golang v0.0.0-20230808230432-28477cd77336
+	github.com/kentik/dashboard-api-golang v0.0.0-20230822181606-6c9a280a9428
 	github.com/kentik/go-metrics v0.0.0-20200109025407-4bfd4a9b42c5
 	github.com/kentik/patricia v0.0.0-20210909164817-21603333b70e
 	github.com/liamg/furious v0.0.0-20191231090757-c295c872d6c1
--- a/go.sum
+++ b/go.sum
@@ -274,8 +274,8 @@ github.com/karrick/godirwalk v1.8.0/go.mod h1:H5KPZjojv4lE+QYImBI8xVtrBRgYrIVsaR
 github.com/karrick/godirwalk v1.10.3/go.mod h1:RoGL9dQei4vP9ilrpETWE8CLOZ1kiN0LhBygSwrAsHA=
 github.com/kentik/api-schema-public v0.0.0-20220322181339-896729e59945 h1:PCi9E7y4j28fqFMjjAVn97jZ90FpbavvbhKa/FGiDZA=
 github.com/kentik/api-schema-public v0.0.0-20220322181339-896729e59945/go.mod h1:3eRiIiOwdOkInkMptcgaG5q4AvDjQd7zvQYGM7KeSzc=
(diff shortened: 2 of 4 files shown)


thezackm:
>QQ -- the device_status is getting complicated. Open to other ideas here on how to manage. Idea being that there are device status (the APs/switches) and the device's clients status (things connected to them).

I assume this covers the former; do we predict a need to monitor clients directly? At scale I'm curious what kind of timeout situations we'd run into there. 

Do you have an example API endpoint for the device's clients status that you're considering?

i3149:
>| do we predict a need to monitor clients directly? At scale I'm curious what kind of timeout situations we'd run into there.

Hard to say this without talking to users, but if I had to guess it would be more a top talkers situation than every single client. EG, https://developer.cisco.com/meraki/api-v1/get-organization-summary-top-clients-by-usage/ Also https://developer.cisco.com/meraki/api-v1/get-organization-clients-overview/

>| Do you have an example API endpoint for the device's clients status that you're considering? I only know of these 3:

No. And I worry that these device and network level calls you list will not work. 

>| Do customers want to know how many clients are connected to a device over time for capacity planning, or are they really interested in specific details like this?

Again we need to talk to customers to find out but my hunch is no. I think that generally corp networks are over provisioned these days and no one really cares about capacity at this level. (True in the US but maybe not in other countries???) Possibly they do care about latency and loss? There is this call https://developer.cisco.com/meraki/api-v1/get-organization-devices-uplinks-loss-and-latency/ . We could expose as a separate option?

>| ktranslate.meraki.clients%

We're trying to but I don't think it works for anyone with a large deployment.

dfaught12345:
Speaking for my case, we are not interested in collecting client data at all at this time.  I do not think that client data collection through the Meraki API would ever reasonably scale up for our prod network, and if we really need it, it is still available in the Meraki Dashboard.  We are primarily looking for exception data for Meraki devices themselves, which we are actually gathering through alert webhooks defined in the Meraki Dashboard, but also continuous measurements that reflect those same exceptions to facilitate availability reporting.

thezackm (approved):

i3149:
These changes are working for @dfaught12345 so we're gonna commit them. Then we'll circle back with new calls for broader support.

PR #607, opened 2023-09-11
Make rpm

i3149:
Hack to ensure that libpcap comes in correctly as a dependency on package build. 

Also removing snmp profile install stage which isn't needed since these come in via git after install now.


thezackm (approved):

ktranslate-package.yml
```diff
@@ -3,19 +3,15 @@ meta:
   vendor: Kentik
   maintainer: Kentik
 
-depends:
-  - git-core
-
-# All fields above marked as `overridable` can be overridden for a given package format in this section.
-overrides:
-  deb:
+conditional:
```
thezackm, line 6: 🥇

thezackm (approved):

PR #613, opened 2023-09-24
Fast ping option.

### LINKED ISSUE #612: FR: Add `fast_poll` option for failed ping checks
As a user, I need to quickly validate potential outages on my devices. In the current setup, the `ping` tests are scheduled on the same interval as the `poll_time_sec`. This means that I potentially need to wait 5+ minutes to validate a "down" device.

Request that functionality is added that follows the following logic:

> When packet loss = 100%, initiate 'fast polling' where we continue to send ping tests every 10s, for a total of 120s to validate a down condition.

This should only apply when `response_time: true | ping_only: true`. A nice to have would be the ability to control the interval and overall duration as well.

i3149:
This creates a global config var `fast_poll` which defaults to false. If true, a fast ping will start when ktrans sees 100% packet loss during a normal ping. It will continue for 120 sec, pinging every 10s or until non total packet loss is seen. 

Use the env var `KENTIK_FAST_PING_DURATION_SEC` to change the default duration of 120s
Use the env var `KENTIK_FAST_PING_TICK_SEC` to change the default ping time of every 10s.

This also changes the underlaying ping library from github.com/go-ping/ping to github.com/prometheus-community/pro-bing. go-ping is deprecated and it's recommended to use the successor library with the same code semantics os pro-bing. 

Closes #612


thezackm (commented):
@i3149 - does this change to the underlying module remove the need for `KENTIK_PING_PRIV`? a quick glance at the new module's readme and I'm not seeing the option for unprivileged UDP packets here.

i3149:
> @i3149 - does this change to the underlying module remove the need for `KENTIK_PING_PRIV`? a quick glance at the new module's readme and I'm not seeing the option for unprivileged UDP packets here.

It does not. See https://github.com/prometheus-community/pro-bing#supported-operating-systems. 

with `KENTIK_PING_PRIV=true` or `KENTIK_PING_PRIV` unset ping still sends ICMP packets while `KENTIK_PING_PRIV=false` sends UDP. 

Because this is a big change I've deployed to the staging branch also for easier testing. Try it out with the `kentik/ktranslate:staging` docker image.

thezackm (approved):

PR #619, opened 2023-10-17
Allow passing meraki api key via env var.

### LINKED ISSUE #618: 💡 FR - Add env var support to Meraki API Key
As a user, I would like to pass an environment variable into the configuration of my Meraki API polling integration to prevent the need of putting my API key in plain text on the YAML file.

i3149:
Use the env var `KENTIK_MERAKI_API_KEY` to optionally load in a Meraki API Key for ktranslate. This value will override anything set in the yaml config. When set, the yaml config can skip the `api_key` setting. 

Closes #618.


thezackm (approved):
🚀

PR #640, opened 2023-12-07
Allow CONFIG-DIR build-arg for building custom configs

### LINKED ISSUE #638: Add build arg for external snmp-base.yaml
I'm maybe missing something, but I don't see an option for copying a local `snmp-base.yaml` file in the Dockerfile.

I'm building the image for Kubernetes (not running locally in Docker) and want to be able to package my config with it during the build process instead of the runtime setting as is done in the [quickstart](https://github.com/kentik/ktranslate/wiki/SNMP-Quickstart#example-output-to-new-relic-sink).

i3149:
Closes #638 

Some people want to specify an alternative config dir. This lets them override with `--build-arg CONFIG-DIR=my-new-dir`


thezackm (approved):
🚀

PR #642, opened 2023-12-12
Adding a retry for 429 errors in common meraki apis

i3149:
See if we hit a 429 code and retry for up to 5 times before giving up.

### DIFF
--- a/pkg/inputs/snmp/x/meraki/meraki.go
+++ b/pkg/inputs/snmp/x/meraki/meraki.go
@@ -54,6 +54,7 @@ const (
 	ControllerKey       = "meraki_controller_name"
 	MerakiApiKey        = "KENTIK_MERAKI_API_KEY"
 	DeviceCacheDuration = time.Duration(24) * time.Hour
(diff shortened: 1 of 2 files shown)


thezackm (dismissed):
Do we know the impact of refreshing a 429 6 times over ~15sec? Wondering if there's a risk of throttling from Meraki?

Otherwise it looks great!

i3149:
So this is why the golang API is annoying. I have an open [issue](https://github.com/meraki/openapi/issues/52) with the swagger spec to expose the header which tells you when to retry. But it hasn't been looked at since August /shrug. Without this header, you're kinda just guessing on how long to wait for. Python defaults to a random number between 1 and 60 sec to sleep. Possibly we should do this? Agree, 5 is kinda a lot of retries, maybe make this 2?

thezackm:
What about a hard cap? User-configurable 0-10 (default: 2) at 5 sec interval?

That would max at 55 seconds for folks trying the 60s poll interval.

i3149:
Deal, will change over to this logic.

thezackm (approved):

PR #645, opened 2023-12-18
Big update to allow ifspeed for slow interfaces. 

### LINKED ISSUE #643: FR: Calculation of IfInUtilization and IfOutUtilization using items collected by SNMP MIB-2 Interfaces
The current IfInUtilization and IfOutUtilization calculations are based on ifHighSpeed (Mbps), which can be obtained from OID 1.3.6.1.2.1.31.1.1.15.

https://github.com/kentik/ktranslate/blob/e5c428c4c4e7949a9d0371de2ffcfa077c461dd6/pkg/formats/nrm/nrm.go#L621-L635

There are cases where iHighSpeed is not supported for older devices or slower interfaces and therefore cannot be calculated.
Could you please consider making it possible to calculate IfInUtilization and IfOutUtilization using items that can be obtained from SNMP MIB-2 Interfaces?

Thank you.

i3149:
Closes #643 

Look at supporting ifSpeed when we need to for interface speed. Need to think through what can go wrong here.

### DIFF
--- a/pkg/formats/ddog/ddog.go
+++ b/pkg/formats/ddog/ddog.go
@@ -412,52 +412,50 @@ func (f *DDogFormat) setRates(ms *datadogV2.MetricPayload, direction string, in 
 	totalPkts := in.CustomBigInt[fmt.Sprintf("ifHC%sUcastPkts", direction)] + in.CustomBigInt[fmt.Sprintf("ifHC%sMulticastPkts", direction)] + in.CustomBigInt[fmt.Sprintf("ifHC%sBroadcastPkts", direction)]
 
 	if ii, ok := f.lastMetadata[in.DeviceName].InterfaceInfo[port]; ok {
-		if speed, ok := ii["Speed"]; ok {
-			if ispeed, ok := speed.(int64); ok {
-				uptime := in.CustomBigInt["Uptime"]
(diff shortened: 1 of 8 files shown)


i3149:
I think this at least will not break existing behavior. User reported that its working for them in staging so let me know what you think @thezackm

thezackm:
@i3149 - looks good; curious if there would be a situation on these older devices where other "high" metrics `ifHCInOctets`, `ifHCOutOctets`, etc. would be missing as well.

i3149:
> @i3149 - looks good; curious if there would be a situation on these older devices where other "high" metrics `ifHCInOctets`, `ifHCOutOctets`, etc. would be missing as well.

These are actually covered in the profiles. You added these override blocks: https://github.com/kentik/snmp-profiles/blob/main/profiles/kentik_snmp/_general/if32-mib.yml#L21

So that the 32bit counters are reported as `ifHCInOctets` and so on. As far as ktrans thinks, everyone everywhere supports high metrics. Just the annoying speed which isn't.

thezackm (approved):

thezackm (approved):

khara-NewRelic:
Hi, my customer has started monitoring for Fujitsu and Yamaha network devices with custom profiles. Then, they has faced this issue of low interface data derived from the devices. The release covering this issue is very important to support their network operation. When is the release scheduled?

PR #647, opened 2024-01-03
Fix flow ingest

i3149:
Updates goflow to 2.1.0, allows `-nf.source auto` to auto-detect flow type, fixes a bug with how sampling works.


i3149:
Lots o excitement here but this is finally the old system with a pcap test.

thezackm (approved):

thezackm (approved):

PR #653, opened 2024-01-23
Adding otel format for ktrans

### LINKED ISSUE #628: Kentik synthetics output to OTLP format
Our friends over at SumoLogic would love to see us output OTLP formatted data from ktranslate instead of the JSON they are using today. Initially they are interested in putting Kentik synthetics into this format but flow data could be interesting too if the output format supports both.

https://opentelemetry.io/docs/specs/otlp/

i3149:
Closes #628 

Add in a otel format for ktrans.


jryburn (approved):
LGTM

PR #656, opened 2024-01-26
Allowing otel to stdout also for debugging

i3149:
More otel work to make easier to debug. Now defaults to printing to stdout unless you override with `otel.endpoint`.


jryburn (approved):
LGTM

PR #663, opened 2024-02-05
Masking out file operations also

### LINKED ISSUE #659: Capability to pickup new and changes to profiles
Provide a capability to pickup new profiles or changes to profiles without having to restart of containers. Is it possible via volume mounting at present if not can we build the capability into the setup.

i3149:
Closes #659

Fixes a bug where the `USR2` signal wasn't getting properly picked up. 

This allows linux and macos version of ktrans to dynamically reload the snmp config with a `kill -s USR2 60161` for example.

### DIFF
--- a/pkg/inputs/snmp/snmp.go
+++ b/pkg/inputs/snmp/snmp.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"os/signal"
 	"strings"
 	"time"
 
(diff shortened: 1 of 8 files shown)


thezackm (commented):
@i3149 - clarification: will this update explicitly listen for changes to the SNMP profiles directory or does it require the user to submit their own kill signal?

If the latter, what is the documented process to provide that signal to the container?

i3149:
> @i3149 - clarification: will this update explicitly listen for changes to the SNMP profiles directory or does it require the user to submit their own kill signal?
> 
> If the latter, what is the documented process to provide that signal to the container?

Good point. I was thinking the latter where you have to shell into each pod and manually signal the ktrans process. But I see why this is non optimal. Let me hook up an inotify solution where it will automatrically reload on `snmp.yaml` change. Shouldn't be too hard, just have to make sure a discovery doesn't also trigger this event. Long winded way of saying the latter but let me try to make it the former.

i3149:
OK, This PR now will set up a watcher on the `mib_profile_dir` value in `snmp.yaml`. By default this is turned off but can be enabled with 

```yaml
global:
    watch_profile_changes: true
```

When a file or directory is removed or written to, ktrans will reload its configs and profiles. 

NOTE: because of a limitation in the watcher [library](https://github.com/fsnotify/fsnotify?tab=readme-ov-file#are-subdirectories-watched) this is not recursive. So if a profile changes in a sub directory you will also need to edit a top level scratch file to not the change. Since the profiles are changing I would assume this will be possible.

thezackm (approved):

PR #668, opened 2024-02-14
Add meraki logs

i3149:
Adding in https://developer.cisco.com/meraki/api/get-network-appliance-security-events/ and https://developer.cisco.com/meraki/api/get-network-events/ to meraki. 

Enable with

```
         preferences:
                log_network_security_events: true
                log_network_events: true
```

log_network_events doesn't seem to have a start time component so it only is called at startup. log_network_security_events is polled each duration. 

Since these are network level calls, be careful about adding lots of networks to the list because of usage limits. 

What do you think?

### DIFF
--- a/pkg/inputs/snmp/x/meraki/meraki.go
+++ b/pkg/inputs/snmp/x/meraki/meraki.go
@@ -359,8 +359,8 @@ func (c *MerakiClient) getNetworkApplianceSecurityEvents(dur time.Duration) erro
 	startTime := time.Now().Add(-1 * dur)
 	startTimeStr := fmt.Sprintf("%v", startTime.Unix())
 
-	var getNetworkEvents func(nextToken string, network networkDesc) error
-	getNetworkEvents = func(nextToken string, network networkDesc) error {
+	var getNetworkEvents func(nextToken string, network networkDesc, timeouts int) error
+	getNetworkEvents = func(nextToken string, network networkDesc, timeouts int) error {
 		params := appliance.NewGetNetworkApplianceSecurityEventsParamsWithTimeout(c.timeout)
 		params.SetNetworkID(network.ID)
 		params.SetT0(&startTimeStr)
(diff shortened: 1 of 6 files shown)


pkg/inputs/snmp/x/meraki/meraki.go
```diff
@@ -319,13 +329,164 @@ func (c *MerakiClient) Run(ctx context.Context, dur time.Duration) {
...
+
+	nextPageEndAt := "" // Start from the very beginning.
+
+	// Check once on startup.
+	err, np := c.getNetworkEvents(nextPageEndAt)
+	if err != nil {
+		c.log.Errorf("Cannot get network events: %v", err)
+	} else {
+		nextPageEndAt = np
+	}
```
thezackm, line 404: could you use the [setPageStartAt](https://github.com/meraki/dashboard-api-go/blob/main/client/docs/GetNetworkEvents200Response.md#setpagestartat) parameter to create a poll interval? 

grabbing the previous [getPageEndAt](https://github.com/meraki/dashboard-api-go/blob/main/client/docs/GetNetworkEvents200Response.md#getpageendat) value to reference on the next polling cycle?

i3149: I'm a little leary about doing that because I don't see it in the docs here: https://developer.cisco.com/meraki/api/get-network-events/

You're looking at the responce body object but if you look at the request one there's no easy place to put the startAt value.

thezackm: i think the idea would be to use it as a query param: 
Get Network Events - Meraki Dashboard API v1 - Cisco Meraki Developer Hub 2024-02-20 at 4 40 44 PM

i3149: Isn't that what I'm doing here? https://github.com/kentik/ktranslate/pull/668/files#diff-5fec11c4942216760fa250830a33713575b00e08ab3fb77527165a55ae693c32R356

This gets the latest timestamp and then will use this for the `startingAfter` param on the next call. Only problem is that the first time the value is empty so you get everything from the last week.

thezackm: d'oh! ok, let's add this to the staging image and see what it looks like; I think there MIGHT be a use case in the future to add a runtime option for how many minutes to look back on start; but until there's a customer demand I don't think we should worry. There's obviously a risk of duplicate data during an agent upgrade, but it should not be super impactful IMO.

thezackm (approved):

thezackm (approved):

PR #669, opened 2024-02-19
Using device name for the device names, if possible

i3149:
Uses the existing device info cache to pull out actual device names fro vpn and uplink status. Reverts to the current behavior if for any reason the device can't be found.


thezackm (approved):
let's try this on staging first

i3149:
On staging, check it out when you like.

thezackm:
staging testing confirmed! 🚀

PR #671, opened 2024-02-22
Explicit enum for meraki device and vpn status

i3149:
if we can, the mapping here would be ideal:
```
{
  online == 1
  alerting == 2
  offline == 3
  dormant == 0
}
```


thezackm (approved):
🚀

PR #695, opened 2024-04-17
Adding an option to support tls in otel

i3149:
Adds in three flags:

```
-otel.tls_cert=certs/cert.pem  -otel.tls_key=certs/key.pem -otel.root_ca=confs/rootCA.crt
```

If all 3 are set, a TLS config is loaded and applied to any otel http and grpc based connections. 

Also bumps the version of otelhttp from 0.45 to 1.25.


jryburn (approved):
LGTM

PR #698, opened 2024-04-22
Allow otel to support snmp traps. 

i3149:
I think this is good now. Tested with:

```
OTEL_METRIC_EXPORT_INTERVAL=30000 KENTIK_PING_PRIV=false ./bin/ktranslate -snmp snmp.yaml -format otel -otel.protocol grpc -otel.endpoint http://localhost:4317/
```

And otel config of:
```yaml
pye ~/tmp $ cat contrib.yaml 
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4222

exporters:
    file:
      path: ./filename.json

service:
  extensions: []
  pipelines:
    metrics:
      receivers: [otlp]
      processors: []
      exporters: [file]
    logs:
      receivers: [otlp]
      processors: []
      exporters: [file]  
```

Both metrics and trap events were received.


i3149:
I'm using slog to format the received traps. End result looks like:

```json
{
  "resourceLogs": [
    {
      "resource": {
        "attributes": [
          {
            "key": "service.name",
            "value": {
              "stringValue": "ktranslate"
            }
          }
        ]
      },
      "scopeLogs": [
        {
          "scope": {
            "name": "github.com/agoda-com/otelslog",
            "version": "0.0.1"
          },
          "logRecords": [
            {
              "timeUnixNano": "1714024418038151000",
              "observedTimeUnixNano": "1714024418038151000",
              "severityNumber": 9,
              "severityText": "INFO",
              "body": {
                "stringValue": "New Trap Event"
              },
              "attributes": [
                {
                  "key": "instrumentation.name",
                  "value": {
                    "stringValue": "snmp-trap-events"
                  }
                },
                {
                  "key": "tags.location",
                  "value": {
                    "stringValue": "$SysDescr"
                  }
                },
                {
                  "key": "provider",
                  "value": {
                    "stringValue": "kentik-trap-device"
                  }
                },
                {
                  "key": "tags.test",
                  "value": {
                    "stringValue": "one"
                  }
                },
                {
                  "key": "src_addr",
                  "value": {
                    "stringValue": "127.0.0.1"
                  }
                },
                {
                  "key": "message",
                  "value": {
                    "stringValue": "ddd~"
                  }
                },
                {
                  "key": "instrumentation.provider",
                  "value": {
                    "stringValue": "kentik"
                  }
                },
                {
                  "key": "TrapName",
                  "value": {
                    "stringValue": "bigipServiceDown"
                  }
                },
                {
                  "key": "TrapOID",
                  "value": {
                    "stringValue": ".1.3.6.1.4.1.3375.2.4.0.10"
                  }
                },
                {
                  "key": "collector.name",
                  "value": {
                    "stringValue": "ktranslate"
                  }
                },
                {
                  "key": "device_name",
                  "value": {
                    "stringValue": "127.0.0.1"
                  }
                },
                {
                  "key": "eventType",
                  "value": {
                    "stringValue": "KSnmpTrap"
                  }
                }
              ],
              "traceId": "",
              "spanId": ""
            }
          ],
          "schemaUrl": "https://opentelemetry.io/schemas/1.21.0"
        }
      ]
    }
  ]
}
```

jryburn (approved):
Code looks good but I would like to clarify the architecture. Is the idea that ktranslate accepts SNMP traps and turns that into Otel events or that it accepts syslog and turns that in Otel SNMP traps or something else?

The reason I ask is that we have a customer, Telxius, who wants to receive "traps from Kentik" when a DDoS alerting policy fires off so they can integrate with Netcool the way their existing DDoS detection system does. I am thinking we might be able to build a creative solution where our notification system sends a webhook or syslog to ktranslate that turns that into an SNMP trap that is sent to Netcool. We could just run ktranslate on their cluster and it would look to them like we are sending traps.

i3149:
> Is the idea that ktranslate accepts SNMP traps and turns that into Otel events or that it accepts syslog and turns that in Otel SNMP traps or something else?

The idea is that ktranslate accepts all SNMP traps like normal. When the otel format is selected, it will output a log to the otel collector as specified on the command line. 

For the Telxius use case I'll flip things around and make ktrans be able to turn the webhook from kentik alerting into a trap send. We'll have to make up some oids for this but shouldn't be too hard.

PR #703, opened 2024-04-26
Let traps have a regex conversion

### LINKED ISSUE #700: Regex capture mapping
In certain scenarios a 1-to-1 mapping of trap attributes is not sufficient. Sometimes the information needed to properly process the trap event at its destination (eg. New Relic) is "hidden" within the attribute value.
A specific example are traps coming from VMWare vCenter. The source of this trap will always be the IP assigned to the vCenter itself, but the true origin of the trap is mentioned within one of the trap attributes. 

To extract the eg. origin, a regex-capture functionality would be extremely valuable. That way one could define which part of an attribute to map to a specific event name. The config could look something like this:

`traps:
  - trap_oid: 1.3.6.1.4.1.41263.999
    trap_name: ntxAlert
    drop_undefined: false
    events:
      - name: ntxAlertCreationTime
        OID: 1.3.6.1.4.1.41263.999.1
      - name: ntxAlertDisplayMsg
        OID: 1.3.6.1.4.1.41263.999.2
      - name: ntxAlertTitle
        OID: 1.3.6.1.4.1.41263.999.3
        regex:
            pattern: /([a-z]*):\s([0-9]-[0-9]-[0-9])\s([a-z]*)/
            capture: Problem with host-id $2`

In the above case, the OID 1.3.6.1.4.1.41263.999.3 contain a text like this:
`There is an issue with: 992-227-332 . The cpu is too high`
The only information we are interested in mapping to the ntxAlertTitle is "Problem with host-id 992-227-332".

i3149:
Closes #700 

(wow, number 700)!

What do you think about this? It uses the existing regex part of snmp profiles so just needed an update to the trap handler. Given an input of 

```
snmptrap -c public -v 2c 127.0.0.1:1620 "" 1.3.6.1.4.1.41263.999 1.3.6.1.4.1.41263.999.3 s "There is an issue with: 992-227-332 . The cpu is too high"
```

Result is `"Problem with.hostid":"992-227-332",` and fully:
```json
[{"src_addr":"127.0.0.1","eventType":"KSnmpTrap","TrapOID":".1.3.6.1.4.1.41263.999","instrumentation.provider":"kentik","instrumentation.name":"netflow-events","collector.name":"ktranslate","TrapName":"ntxAlert","Problem with.hostid":"992-227-332","device_name":"127.0.0.1","provider":"kentik-trap-device"}]
```

The trap.yaml file is extended with:

```yaml
  - trap_oid: 1.3.6.1.4.1.41263.999
    trap_name: ntxAlert
    drop_undefined: false
    events:
      - name: ntxAlertCreationTime
        OID: 1.3.6.1.4.1.41263.999.1
      - name: ntxAlertDisplayMsg
        OID: 1.3.6.1.4.1.41263.999.2
      - name: ntxAlertTitle
        OID: 1.3.6.1.4.1.41263.999.3
        conversion: "regexp:\\:\\s(?P<hostid>[0-9]+-[0-9]+-[0-9]+)\\s"
        tag: "Problem with"
```

`conversion` sets up a capturing regex while the `tag` supplies the key part of the key-value pair. 

This relies on [named captures](https://github.com/StefanSchroeder/Golang-Regex-Tutorial/blob/master/01-chapter2.markdown#named-matches) to set the key part of the map.

### DIFF
--- a/pkg/inputs/snmp/traps/traps.go
+++ b/pkg/inputs/snmp/traps/traps.go
@@ -245,15 +245,23 @@ func (s *SnmpTrap) handle(packet *gosnmp.SnmpPacket, addr *net.UDPAddr) {
 		case gosnmp.OctetString:
 			if value, ok := snmp_util.ReadOctetString(v, snmp_util.NO_TRUNCATE); ok {
 				if res != nil && res.Conversion != "" { // Adjust for any hard coded values here.
-					_, value, _ = snmp_util.GetFromConv(v, res.Conversion, s.log)
-				}
-				if !utf8.ValidString(value) { // Print out value as a hex string.
-					value = fmt.Sprintf("%x", v.Value)
-				}
-				if res != nil {
-					dst.CustomStr[res.GetName()] = value
-				} else {
-					dst.CustomStr[v.Name] = value
+					_, sval, mval := snmp_util.GetFromConv(v, res.Conversion, s.log)
+					if len(mval) > 0 { // List the regex matches.
+						for k, v := range mval {
+							dst.CustomStr[k] = v
+						}
+					} else {
+						dst.CustomStr[res.GetName()] = sval
+					}
+				} else { // No conversion.
+					if !utf8.ValidString(value) { // Print out value as a hex string.
+						value = fmt.Sprintf("%x", v.Value)
+					}
+					if res != nil {
+						dst.CustomStr[res.GetName()] = value
+					} else {
+						dst.CustomStr[v.Name] = value
+					}
 				}
 			}
 		case gosnmp.Counter64, gosnmp.Counter32, gosnmp.Gauge32, gosnmp.TimeTicks, gosnmp.Uinteger32, gosnmp.Integer:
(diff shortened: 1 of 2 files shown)


thezackm (changes requested):
```
"Problem with.hostid":"992-227-332"
```

The space in that key name is going to be a little problematic and cumbersome for querying later. Can we shift this to use the direct value of `tag` and not add a prefix? 

e.g.;
```yaml
      - name: ntxAlertTitle
        OID: 1.3.6.1.4.1.41263.999.3
        conversion: "regexp:\\:\\s(?P<hostid>[0-9]+-[0-9]+-[0-9]+)\\s"
        tag: host_id
```

```
"host_id":"992-227-332"
```

I don't think the named capture is necessary in this context. Similar to how we setup regex on SNMP polling, the user should be able to provide their key name directly instead of forcing an inference.

i3149:
Good point. Moved to be just the key for the naming.

thezackm (approved):

PR #704, opened 2024-04-29
Send trap as format

i3149:
First pass on code to send any inputs ktrans gets out again as a snmp trap. Uses the generic trap value of `.1.3.6.1.4.1.41263.6169` where 6169 comes from Kentik's AS number.

Tested with

```
./bin/ktranslate -snmp trap-v2.yaml \
  -format snmp \
  -snmp.format.conf trap-v2.yaml
```

Where `trap-v2.yaml` is 

```yaml
trap:
    listen: 127.0.0.1:1620
    community: public
    drop_undefined: false
    version: v2c
    transport: ""
    trap_only: true
    endpoint: 127.0.0.1
    endpoint_port: 1620
global:
  mib_profile_dir: ./config/profiles
  timeout_ms: 3000
```

Supports sending v1, v2c or v3.


i3149:
Tested this with the webhook from kentik and things are good to go.

jryburn (approved):
LGTM. Only comment is does it make sense to hard-code the OID or should we make that an option that can be fed in on the CLI or pulled in via an env file? It's been a LOOONG time since I messed with traps so I'm not sure what the best practices are there.

i3149:
Generally I've seen OIDs hard coded to be specific things. But I'd argue its too early to make a Kentik mib so we can use this as a stepping stone.

PR #708, opened 2024-05-06
Logging poll interval for snmp

### LINKED ISSUE #707: 💡 FR - Decorate Uptime with `poll_interval_sec` attribute
As a user, it's helpful during debugging to know what interval should be expected for my various devices so I can more easily identify possible timeout and/or capacity issues.

Requesting that we decorate `device_poll_time_sec` for the device into its `kentik.snmp.Uptime` metric. We can ignore the singular metrics that have static rates set in their profile (CPU and Memory mostly) and only return the device attribute from the local configuration.

i3149:
Closes #707 

Adds in `SysPollInterval` attribute on Uptime for logging the configured polling interval. 

```
"SysPollInterval": "30s",
```

### DIFF
--- a/pkg/inputs/snmp/metadata/poll.go
+++ b/pkg/inputs/snmp/metadata/poll.go
@@ -2,6 +2,7 @@ package metadata
 
 import (
 	"context"
+	"fmt"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


thezackm (approved):
lgtm

PR #721, opened 2024-07-05
Fix 720, allow strings in the comparison operator

### LINKED ISSUE #720: 💡 `GetCondition` needs to respect string matching as well as integer values
revisiting PR #350 here...

The [GetCondition](https://github.com/kentik/ktranslate/blob/65e861175b50a96ecffa85ce7659611ed737092c/pkg/inputs/snmp/mibs/profile.go#L950-L968) function looks like it will error if the comparison is anything but an integer right now.

We have a need to expand this functionality to support a string match as well for OIDs that use the `SnmpAdminString` type like this:
```yaml
      - name: cempMemPoolHCUsed
        OID: 1.3.6.1.4.1.9.9.221.1.1.1.1.18
        poll_time_sec: 60
        tag: MemoryUsed
        allow_duplicate: true
        condition: cempMemPoolName="DP System memory"
```

i3149:
Closes #720 

Does this work for you? I've just updated the comparison operator to handle strings and ints. Also some more unit tests. Tested cases:

```
		`cempMemPoolName="DP System memory"`: []condTest{
			condTest{si: "ddd", r: false},
			condTest{si: "DP System memory", r: true},
		},
		`MemoryCache=4`: []condTest{
			condTest{ii: 2, r: false},
			condTest{ii: 4, r: true},
		},
```

### DIFF
--- a/pkg/inputs/snmp/metrics/device_metrics.go
+++ b/pkg/inputs/snmp/metrics/device_metrics.go
@@ -480,7 +480,7 @@ func (w wrapper) checkCondition(idx string, results []wrapper) bool { // Check c
 			oidNameWr := wr.mib.GetName() // Does the name match our target?
 			if oidNameWr == w.mib.Condition.TargetName {
 				// If it does, does it match our target value?
-				return snmp_util.ToInt64(wr.variable.Value) == w.mib.Condition.TargetValue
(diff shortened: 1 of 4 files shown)


thezackm (approved):
LGTM

PR #727, opened 2024-07-23
Fix package build for centos 8, remove centos 7.

i3149:
Deprecate centos 7, fix centos 8.


thezackm (approved):

thezackm:
@i3149 - clarification - you're deprecating centos 8 too, or simply moving to the redhat universal base image for testing and a customer can still deploy on centos 8 with support going forward?

PR #730, opened 2024-07-27
Bug with filters not reseting kentik flow. 

i3149:
Fixing a bug where filters which remove a lot of traffic will eventually block any new data. So filters will work for a bit and then eventually no new data will get past the filters. Fixed with a one liner but several weeks to track down.


jryburn (approved):
LGTM

PR #731, opened 2024-08-02
Updates to rollups and tag map

i3149:
Few updates here:

1) the tag_map file now can contain functions which modify text directly. For example, `101,DST_SUBSCRIBER_ID,to_hex`.

2) To get all values of a rollup, set `-rollup_top_k="-1"` The -1 value means return everything.  

3) Parquet format now gets the `.parquet` suffix added.


jryburn (approved):
LGTM

PR #734, opened 2024-08-10
Guard on extra packets being recv in a ping interval

### LINKED ISSUE #732: Poller picking up false packet loss on ping_only tests
We are using ktranslate agent running as a docker container with NewRelic to do some **ping_only** checks.  We know the version v2.1.23 from Feb2024 is working as expected and returning the expected results.  However any of the recent versions (even v2.1.26 from May2024) are producing intermittent packet loss metrics.  If I had to make a guess it seems v2.1.25 is where something changed related to ping - https://github.com/kentik/ktranslate/releases/tag/v2.1.25 

Based on the NewRelic data collected it appears there were occasionally 6 packets were sent instead of the usual 5. Packets received remains as 5 and its shows as packetLoss.  Let me know if I can provide additional info around this issue.

OcassionalPacketSent6insteadof5_PacketLossIssue
NR_NetworkPollerPacketLossIssue

i3149:
Closes #732 

Adds a guard so if there's an extra packet seen across ping intervals we don't report spurious packet loss.

### DIFF
--- a/pkg/inputs/snmp/metrics/device_metrics.go
+++ b/pkg/inputs/snmp/metrics/device_metrics.go
@@ -19,8 +19,9 @@ import (
 )
 
 type pingStatus struct {
-	sent     uint64
-	received uint64
+	sent         uint64
+	received     uint64
+	lastDiffSent uint64
 }
 
 type DeviceMetrics struct {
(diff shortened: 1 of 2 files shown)


jryburn (dismissed):
LGTM

i3149:
One more time, sorry I think I found a better way.

jryburn (approved):
Definitely seems cleaner than the previous approach. LGTM

PR #737, opened 2024-08-15
Fixing too many pings one more time

### LINKED ISSUE #736: Reopening packet loss on ping_only tests from issue732
This is a follow-up to issue 732 - https://github.com/kentik/ktranslate/issues/732 

Pulled version (-- Version kt-2024-08-12-10360414534; Build Mon Aug 12 22:39:38 UTC 2024) to test and did not notice any changes in the results.  Same results were obtained where 6 packets sent and 5 received intermittently.

Pulled version via:
docker pull quay.io/kentik/ktranslate:v2

If the fix is not in the above release please provide the version info that I should be pulling to test.

i3149:
Closes #736 

Adding an explicit guard for the too many ping bug.

### DIFF
--- a/pkg/inputs/snmp/metrics/device_metrics.go
+++ b/pkg/inputs/snmp/metrics/device_metrics.go
@@ -27,6 +27,7 @@ type DeviceMetrics struct {
 	profileName string
 	oids        map[string]*kt.Mib
 	missing     map[string]bool
+	lastSent    int
 }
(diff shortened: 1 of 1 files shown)


i3149:
Was able to reproduce and validated this one is working so I'm confident here.

jryburn (approved):
LGTM

PR #739, opened 2024-08-23
Moving to using a home grown ping engine.

### LINKED ISSUE #738: Reopening packet loss on ping_only tests from issues 732 and 736
This is a follow-up to issues 732 and 736 - https://github.com/kentik/ktranslate/issues/736

There was still packetloss recorded on the initial polling but the bigger issue in the latest version released was some of the devices did not get polled for a lengthly time period as in below example.   The poller was started up at 3pm but it seems this specific host got polled in the beginning and then no more polls on it until like 5:15pm timeframe.  So a >2 hour gap.  There were other hosts having the same type of behavior as well.

image

i3149:
Switching up to a new ping system because the other one was too buggy. This is mostly a drop in replacement but is a lot simpler. 

Also bumping go version to 1.22. 
Closes #738

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -108,7 +108,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v2
         with:
-          go-version: '1.21.6'
+          go-version: '1.22.6'
 
       - name: Get Deps
         shell: bash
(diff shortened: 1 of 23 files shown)


jryburn (approved):
LGTM

jryburn:
This is attempting to address issue #738

PR #744, opened 2024-09-06
Adding a test for upper case OR and AND in filters

i3149:
This fixes a bug @kentik-rbarnes found where `and` works for filters now but `AND` doesn't.  Also some more tests on this.


jryburn (approved):
LGTM

PR #745, opened 2024-09-07
Adding interface provider to firehose flow

i3149:
This adds in Interface Provider, SiteTitle and SiteCountry to the flow exported from kentik via firehose. Data was present in the API before but not dumped out.


jryburn (approved):
LGTM

PR #747, opened 2024-09-09
Fixing interface api

i3149:
The way the kentik API works for interfaces moved. This updates the API code to correctly take in the needed data. 

Now we can correctly see data like `"input_provider":"MICROSOFT`


jryburn (approved):
LGTM

PR #748, opened 2024-09-09
removing windows package because no one uses it

i3149:
I don't think there are any current windows users and this build just broke. Removing for now because its not worth the effort to fix.


jryburn (approved):
Makes sense to me. We don't support windows for any other Kentik built exec binaries.

PR #749, opened 2024-09-10
Allowing rollups to keep undefined terms if needed

i3149:
Adds in a new flag `-rollup_keep_undefined` which will not remove undefined rollup values but instead replace them with the string `undefined`. This is a customer request.


jryburn (approved):
LGTM

PR #751, opened 2024-09-19
Fixing bug where the primary dim might be not the first one

i3149:
Adds a few tests for rollups. Also fixing a bug where for a definition like 

```
-rollups s_sum,sum_bytes_in,in_bytes,custom_str.output_provider,device_name,output_int_desc,output_int_alias
```

You get a result like

```
"dimension": "device_name^outout_int_desc^output_int_alias^output_provider",
```

@kentik-rbarnes this fixes the bug you raised.


jryburn (approved):
LGTM

PR #755, opened 2024-09-28
Allow named filters to let multiple rollups hang off one instance

### LINKED ISSUE #753: Firehose support for multiple filter and rollups in a single instance
**What are you trying to achieve?**

Combining multiple sets of filter/rollup combinations to a single instance of firehose caused unexpected data outputs. Due to this it was necessary to tee flow from one instance of firehose to a second instance allowing each instance to utilize it's own filter set and output rollups as required.

The use of two instances effectively doubles the expected CPU/Memory required to implement the solution
If each firehose instance is limited to only a single --filter then any additional rollups that might require a slightly different filter set would require yet another instance to be tee'd and instantiated increasing the footprint further

The request here is to implement a **single firehose instance** that can take multiple **sets of filter criteria and matching rollup criteria** optimizing for resources to limit the required compute footprint required.

**What is your current work around? (optional)**

tee and spawn a new instance of firehose for each --filter --rollup pair required for reporting purposes

cc @kentik-rbarnes please provide examples of sets **filter + rollup** that can be used for testing.

i3149:
Closes #753 

Lets filters and rollups work together so that you can attach a named filter to a given set of rollups. 

```
   -filters "string,custom_str.dst_network_bndry,==,external,sum_external"
        -rollups s_sum,sum_external,in_bytes,custom_str.src_subscriber_id
```

For example.

### DIFF
--- a/pkg/cat/kkc.go
+++ b/pkg/cat/kkc.go
@@ -121,6 +121,7 @@ func NewKTranslate(config *ktranslate.Config, log logger.ContextL, registry go_m
 	for _, filter := range filters {
 		if filter.GetName() == "" { // No name means a global application.
 			fullSet = append(fullSet, filter)
+			continue
 		}
 
 		found := false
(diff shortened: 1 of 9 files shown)


jryburn (approved):
LGTM

PR #758, opened 2024-10-03
Adding has filter logic for strings

i3149:
Adding in has and notHas filters for strings. Before, there wasn't a good way to test if a given filter key is present regardless of value. Operators are `=~` and `!~`. For example:

```
		"string,custom_str.foo,=~,",       // This contains the key custom_str.foo with any value. 
		"string,custom_str.foo,!~,",        // This returns true if the key custom_str.foo is not set. 
```


jryburn (approved):
LGTM

PR #760, opened 2024-10-10
Adding in ult exit data when present

### LINKED ISSUE #759: Firehose Support for Ultimate Exit Dimensions
Current FH export is limited to peering_edge (inbound from external classified interfaces, outbound from backbone classified interfaces) if we want to report on outbound traffic by Peer/Provider.

Ideal spot to make this measurement for outbound traffic is actually closer to the access layer (ie. subscribers) where traffic density is somewhat lower (vs a multi Tbit backbone link) which provides better sample rate fidelity for customer traffic. For this to work - we'll need to also include UE dimensions as part of firehose export so we can sample closest to customer, but still report on which providers that customer is utilizing.

i3149:
This will add in `ult_exit_port_alias` and `ult_exit_port_description` when Ultimate Exit data is present for a given device. 

Closes #759

### DIFF
--- a/cmd/ktranslate/main.go
+++ b/cmd/ktranslate/main.go
@@ -72,7 +72,7 @@ func init() {
 	flag.IntVar(&threads, "threads", 1, "Number of threads to run for processing")
 	flag.IntVar(&threadsInput, "input_threads", 1, "Number of threads to run for input processing")
 	flag.IntVar(&maxThreads, "max_threads", 1, "Dynamically grow threads up to this number")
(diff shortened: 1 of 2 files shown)


jryburn (dismissed):
LGTM but I can't speak to if it fully solves @kentik-rbarnes problem statement.

i3149:
One more set of 👀 plz. I think we're good now with @kentik-rbarnes.

jryburn (approved):
LGTM

PR #762, opened 2024-10-16
first pass on docs to send data to grafana cloud

i3149:
Working on examples of integrations into grafana cloud.

### DIFF
--- a/hack/grafana-cloud/config.alloy
+++ b/hack/grafana-cloud/config.alloy
@@ -1,0 +1,189 @@
+prometheus.exporter.self "alloy_check" { }
+
+discovery.relabel "alloy_check" {
+  targets = prometheus.exporter.self.alloy_check.targets
+
+  rule {
+    target_label = "instance"
+    replacement  = constants.hostname
+  }
+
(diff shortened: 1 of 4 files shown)


hack/grafana-cloud/config.alloy
jryburn, line 1: We probably only need the metrics portion of this, right? Is the logs and traces in here to future proof it in case we decide to add additional functionality?

jryburn (approved):
Makes sense and I like this approach much better than using Loki as Grafana had suggested to Erik in the meeting. See my comment on the config.alloy file though.

hack/grafana-cloud/config.alloy
i3149: Its kinda an open question. From my further experiments, there's no way to put 4k FPS flow into prometheus as metrics. We'd have to send these in as top K rollups. What you can do is send as logs to loki but after an hour this exhausted the 50GB of free plan data.

i3149:
Since this is all otel this in theory should work well with sumo also but I haven't tested this.

PR #765, opened 2024-10-17
Update Otel to Support Rollups

i3149:
This updates the otel format to support rollups of data. 

Also changes how flows by default are sent on, from a metric to a log. I'm finding that collectors can't handle the volume of flow as metrics but work OK as logs.


jryburn (approved):
LGTM

PR #769, opened 2024-10-25
Allow dimensions to be joined together in rollups.

i3149:
We have a customer request to allow rollups to merge 2 dimensions into one. For example, given the definition here:

```
sum,sum_bytes_in,in_bytes,aaa$$---$$bbb,ccc
```

dimensions aaa and bbb would be merged together into aaa---bbb. The `$$` token is used to wrap the inner join token, which in the above example is `---`. 

@kentik-rbarnes , your thoughts too? 

NOTE: this version only supports grouping 2 dimensions into 1. Is this OK or do we need to be able to merge N dimensions into 1?

### DIFF
--- a/pkg/rollup/rollup.go
+++ b/pkg/rollup/rollup.go
@@ -27,6 +27,7 @@ const (
 
 	KENTIK_EVENT_TYPE = "KFlow:%s:%s"
 	UndefinedKey      = "undefined"
+	DimJoinToken      = "$$"
 )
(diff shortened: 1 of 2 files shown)


kentik-rbarnes:
so far we've only been asked to merge two dims so I think this is sufficient. until we get asked to merge three!

jryburn (approved):
LGTM

PR #771, opened 2024-10-28
No buffer mode for parquet files which can not be appended to

i3149:
Fixes a bug where parquet files will get corrupted because you can't just append one to another.

Adds a new type of data option which forces sinks to flush the data right away instead of buffering. 

@kentik-rbarnes this is the fix you're waiting on.


jryburn (approved):
LGTM

PR #773, opened 2024-11-05
fixing a case where some oid values do not get matched in disco

i3149:
For an oid like `SNMPv2-MIB::sysObjectID.0 = OID: SNMPv2-SMI::enterprises.119.1.84.16.1` disco wasn't matching correctly. This should parse it as a string. 

Causes the disco to add the device but not pull in the correct profile.


jryburn (approved):
LGTM

PR #775, opened 2024-11-14
Moving AWS s3 api to v2

i3149:
Work in progress to move the s3 apis to v2. Still need to test all of the branches and make sure its a good idea.


i3149:
Messy but I think this is ready to be looked at. trying to make a way to remove dangling $folder$ files which get dropped into s3 likely by a 3rd party reader.

jryburn (approved):
LGTM

PR #782, opened 2024-12-24
Adding a first ping so that its not 2x the period to get first pings

i3149:
Right now its 2x the ping period to get first results back. This ads an extra ping when things start so its only 1x to get first results into the system.


jryburn (approved):
LGTM

PR #784, opened 2025-01-16
Now adding in service_name value when flow is seen from unkown

i3149:
When processing netflow, ktrans will drop the service_name tag unless the flow is from a known device. This adds in a "missing device" device so there's always a service_name value set.


jryburn (approved):
LGTM

PR #785, opened 2025-01-17
When syslog sends a hostname as ip, try looking it up

i3149:
Sometimes syslog will send a hostname value but not a client value. In this case, try looking up the hostname in the list of snmp discovered devices by ip as a fallback.


jryburn (approved):
LGTM

PR #788, opened 2025-01-31
when there is no ipv6 interface on a host right now ping will crash

i3149:
This making ping a bit smarter, only using ipv6 if we need to ping an ipv6 ip and vice versa.


jryburn (approved):
LGTM

i3149:
Closes https://github.com/kentik/ui-app/issues/26131

PR #789, opened 2025-02-06
re-re doing the build system

i3149:
Another change to the build system to make multi-arch builds work faster and more reliably.


jryburn (approved):
LGTM

PR #793, opened 2025-02-12
Fixing a crash when sockets are not nil

### LINKED ISSUE #794: Ktranslate crashing after commit from #788 - `panic: runtime error`
Customer doesn't have an IPv6 interface on their host and is using `ping_only` devices in their config file. Following commit from #788, container crashes with this error on start-up:

```golang
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1091282]
```

i3149:
Closes #794 
Forgot that sometimes theres a nil and sometimes theres not. 

Internal #11600


jryburn (approved):
LGTM

PR #796, opened 2025-02-26
allowing custom action for elastic format

### LINKED ISSUE #795: Ktranslate - Elasticsearch Update
The current -format=elasticsearch formats the data like this:
```
{ index {} }
{ field1:value, field2:value ...}
{ index {} }
```
This works fine when using the _bulk index endpoint when using index.mode=standard in elasticsearch (for example -http_url 'http://elasticsearch:9200/kentik/_bulk'. However elasticsearch is deprecating rollups, which summarizes data, in favor of downsampling, which only works when you use index.mode=time_series . That mode can only use the method create so I think it's expecting the data to look like this:
```
{ create {} }
{ field1:value, field2:value ...}
{ create {} }
```

It work by changing index on line 21 of https://github.com/kentik/ktranslate/blob/main/pkg/formats/elasticsearch/elasticsearch.go to create and rebuilding it. However, can we update ktranslate and make that field configurable, or at least change it to create so elasticsearch works in the future. Thank you.

i3149:
Closes #795

Lets you set the `--elastic.action create` flag which let you customize the action used for elastic format.

### DIFF
--- a/pkg/formats/elasticsearch/elasticsearch.go
+++ b/pkg/formats/elasticsearch/elasticsearch.go
@@ -4,11 +4,13 @@ import (
 	"bufio"
 	"bytes"
 	"compress/gzip"
+	"flag"
 	"fmt"
 	"io"
 	"strings"
 	"time"
 
+	"github.com/kentik/ktranslate"
 	"github.com/kentik/ktranslate/pkg/formats/util"
 	"github.com/kentik/ktranslate/pkg/kt"
 	"github.com/kentik/ktranslate/pkg/rollup"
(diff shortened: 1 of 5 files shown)


pkg/formats/elasticsearch/elasticsearch.go
```diff
@@ -18,21 +20,31 @@ import (
...
-	actionEntry = `{"index":{}}`
+	actionEntryFormat = `{"%s":{}}`
 )
 
+var (
+	actionEntrySet string
+)
+
+func init() {
+	flag.StringVar(&actionEntrySet, "elastic.action", "index", "Use this action to when sending to elastic.")
```
jryburn, line 31: Typo?

"Use this action to when sending to elastic." should be "Use this action when sending to elastic." I believe.

jryburn (changes requested):
Minor change to help field but otherwise looks good.

pkg/formats/elasticsearch/elasticsearch.go
i3149: good catch!

jryburn (approved):

PR #797, opened 2025-03-12
Update README.md for snmp_discovery_min and snmp_discovery_on_start

khara-NewRelic:
Added in line 186 to 189

  -snmp_discovery_min int
    	If set, run snmp discovery on this interval (in minutes).
  -snmp_discovery_on_start
    	If set, run snmp discovery on application start.


i3149 (approved):
LGTM!

PR #807, opened 2025-04-14
fixing prom when tags present

i3149:
Fixing a dumb thing where tag labels will mess up prom exports.


Mesverrum (approved):
Makes sense to me

PR #809, opened 2025-04-28
Fixing a race condition on tags getting set

i3149:
Adds a lock and some guards to keep complicated tags from getting over written. 

Also adding in a secondary hard check for number of tags. 

`expected_device_tags: 17` in the per device config yaml will force exactly 17 tags on this device or else it will panic with:

```
panic: Wrong number of user tags for device bart found: 13 (expected 17)
```


jryburn (approved):
LGTM

i3149:
One more plz @jryburn , I added a belt and suspenders check for this.

jryburn (approved):
LGTM

PR #811, opened 2025-05-05
Better check for when an ip is v4 or 6 when setting up ping

i3149:
Fixing a reported bug where we don't correct detect if there's an ipv6 interface present.


jryburn (approved):
LGTM

PR #826, opened 2025-06-12
moving to a sucks more syslog

i3149:


i3149:
> Any particular reason to be aware of regarding the switch to the other library?

There's another syslog https://github.com/alistairking/go-syslog which kentik wants to promote to the real slim syslog. But its very different from the og go-syslog so it had to get renamed.

PR #834, opened 2025-07-31
Code to allow custom s3 sink endpoints

i3149:
This adds two new flags, `s3_endpoint` and `s3_signing_region`. These let you override the default s3 regions and use an onprem s3 compatible setup. Tested working with a customer system.


jryburn (approved):
LGTM

PR #837, opened 2025-08-08
Adding a default for otel values so that custom types still get defined

### LINKED ISSUE #836: ktranslate otel format - dropping attributes
**Description**
I am using an otel collector to dual export metrics to 2 backends (new relic and an in-house platform).

```
docker run -d \
  --name ktranslate-otel \
  --restart unless-stopped \
  --user `id -u`:`id -g` \
  -v /opt/newrelic-npm/snmp-otel.yaml:/snmp-base.yaml \
  -e AWS_ACCESS_KEY_ID=123 \
  -e AWS_SECRET_ACCESS_KEY=456 \
  -e AWS_REGION=region\
  kentik/ktranslate:kt-2025-07-22-16455751706 \
  -format=otel \
  -format_metric=otel \
  -otel.protocol=http \
  -otel.endpoint=https://otel-collector:4318 \
  -snmp /snmp-base.yaml \
  -log_level=debug \
  -metrics=jchf \
  -service_name=ktranslate-otel \
  -sinks=otel \
  -tee_logs=true
```

**Issue**
When sending ktranslate otel metrics to New Relic, the provider attribute is getting dropped 
This affects how NR matches the device to its [entity definition](https://github.com/newrelic/entity-definitions/tree/main/entity-types) and how it displays in the UI

**Cause**
The GetTagValues() method has a silent default: case that drops unknown types, line [693](https://github.com/kentik/ktranslate/blob/main/pkg/formats/otel/otel.go).

kt.Provider is a custom type (type Provider string). When the switch statement hits the provider value, it doesn't match any of the cases (string, int64, int32, etc.) and falls into the default: case which silently drops the attribute.

**Proposal**
I've tested casting unknown attribute types to string and it works for kt.Provider 
Can you think of any issues with this approach ?

```
func (d *OtelData) GetTagValues() attribute.Set {
	res := make([]attribute.KeyValue, 0, len(d.Tags))
	for k, v := range d.Tags {
		switch t := v.(type) {
		case string:
			res = append(res, attribute.String(k, t))
		case int64:
			res = append(res, attribute.Int64(k, t))
		case int32:
			res = append(res, attribute.Int64(k, int64(t)))
		case float64:
			res = append(res, attribute.Float64(k, t))
		case uint32:
			res = append(res, attribute.Int64(k, int64(t)))
		case uint64:
			res = append(res, attribute.Int64(k, int64(t)))
		default:
			// Convert unknown types to string representation
			res = append(res, attribute.String(k, fmt.Sprintf("%v", t)))
		}
	}
	s, _ := attribute.NewSetWithFiltered(res, func(kv attribute.KeyValue) bool { return true })
	return s
}
```

i3149:
Closes #836

I forgot that custom types will fall out of this switch. This update will force random types to their string rep.

### DIFF
--- a/pkg/formats/otel/otel.go
+++ b/pkg/formats/otel/otel.go
@@ -691,6 +691,8 @@ func (d *OtelData) GetTagValues() attribute.Set {
 		case uint64:
 			res = append(res, attribute.Int64(k, int64(t)))
 		default:
+			// Convert unknown types to string representation
+			res = append(res, attribute.String(k, fmt.Sprintf("%v", t)))
 		}
 	}
 	s, _ := attribute.NewSetWithFiltered(res, func(kv attribute.KeyValue) bool { return true })


jryburn (approved):
LGTM

Issue #842, opened 2025-09-07
New Relic 'logtype' attribute is not set correctly in batches that have a mix of container logs and syslogs


steventine:
I very much appreciate the great work that was done in https://github.com/kentik/ktranslate/commit/a2f9eb86270bdc897e401dd09b34fd55b83986a3  per https://github.com/kentik/ktranslate/issues/209 to decorate the container logs from ktranslate with the `logtype` of `ktranslate-health` however I was seeing situations where sometimes the logs were decorated (and hence nicely parsed) while other times they weren't.

For example, in New Relic:
```SELECT `container_service`,`plugin.type`,`severity`,`message` FROM Log 
WHERE `collector.name` = 'ktranslate' AND `device_name` IS NULL```

<img width="1756" height="203" alt="Image" src="https://github.com/user-attachments/assets/8c52d201-13a5-4f21-af88-8672ee465ea6" />

(notice the first two are decorated, while the last two aren't)

After lots of searching through code and debugging, I think the problem is that the `logtype` attribute is set as a `Common.Attribute` (applying to the entire batch) and it's only set if **NONE** of the logs in the batch being sent are a `syslog`.  However, in the situation where at least one log in the batch **IS** a syslog, the container logs aren't properly decorated and hence aren't parsed :-(
https://github.com/kentik/ktranslate/blob/f18e2272ced7d0ca045f0d957ea68754a9cca466/pkg/sinks/nr/nr.go#L400-L403

Would it make sense to never mix syslogs and non-syslogs in the same batch since they have to share the `Common.Attribute`?  

Maybe in the code below (which is already copying each log to a new logset) create a syslog_logset and a non_syslog_logset...then send them independently, with the latter getting the `logtype` of `ktranslate-health`?

https://github.com/kentik/ktranslate/blob/f18e2272ced7d0ca045f0d957ea68754a9cca466/pkg/sinks/nr/nr.go#L391-L396

steventine:
Maybe something like https://github.com/steventine/ktranslate/commit/089887e4b009ff16e571b1ec712aa2f00a0011c0 ?

(my apologies as I've never written any go code before so this might be totally wrong, but hopefully it's a conceptual example of the idea I was proposing)

steventine:
Or is there an easy way to apply the `plugin.type` and `logtype` at the individual log level (instead of inside the `Common.Attributes` that applies to all logs in the batch)?

steventine:
[...]

PR #844, opened 2025-09-09
Set New Relic 'logtype' attribute correctly for container logs & syslogs in the same batch 

steventine:
Per the discussion in https://github.com/kentik/ktranslate/issues/842 here is a draft PR that ensures that the `logtype` attribute is correctly set for container logs (aka `ktranslate-health` logs) vs `syslogs`.  This is accomplished by sending the different types of logs in different batch messages as all logs in a batch message share the same set of `Common.Attributes`


i3149 (approved):
LGTM

Issue #835, opened 2025-08-05
ktranslate does not output the correct logTimeStamp on OS with non-UTC timezone settings


qryuu:
### What is the problem the customer is trying to solve?
When ktranslate is run on an OS with a non-UTC setting, ktranslate's own logTimeStamp is shifted by the time difference from UTC.

### Why is it important for users that we fix that problem?
This is a problem that all users who use ktranslate on OSes that are not set to UTC will face. If the logTimeStamp is shifted, it will be difficult to correlate with other logs, and it will be a major obstacle to troubleshooting.

### What is the objective they are pursuing?
The user wants ktranslate to output the correct logTimeStamp even on OSes running in timezones other than UTC.

### Use cases and user stories:
A user in a non-UTC time zone, such as JST, runs ktranslate. The user wants to check the logs of ktranslate, but the logTimeStamp is shifted from the actual time, making it difficult to analyze the log.

### Is there a known workaround?
The user can set the OS's time zone to UTC, but this may affect other applications running on the OS.

qryuu:
Would you mind looking into this issue?

steventine:
@qryuu I had this problem too.  I'm using New Relic and I made this change, which fixed it for me - https://github.com/kentik/ktranslate/pull/845

I'm hopeful we can get a change like this (even if it's behind a new config flag to be safe) into ktranslate

PR #845, opened 2025-09-10
Add time zone to ktranslate logs so host doesn't need to always be in UTC

steventine:
As reported in https://github.com/kentik/ktranslate/issues/835, when ktranslate is run on an OS with a non-UTC setting, ktranslate's own logTimeStamp is shifted by the time difference from UTC.  This is because the format of the logs does not include any TZ information, so the destination (like NewRelic) assumes the timestamp is in UTC...but the timestamp isn't really UTC:
https://github.com/kentik/ktranslate/blob/f18e2272ced7d0ca045f0d957ea68754a9cca466/pkg/util/logger/logger_writer.go#L18

This PR proposes the idea of explicitly adding the current OS's time zone offset for ktranslate logs.  As this format matches ISO8601 it's likely that all the sinks will handle this change, I don't really have a way of knowing.  I've tested it with New Relic and it works.

Example log after this change:
`2025-09-10T08:22:15.997-04:00 ktranslate/home-assistant-ktranslate [Info] baseserver.metaserver Listening on 127.0.0.1:38149`

### DIFF
--- a/pkg/util/logger/logger_writer.go
+++ b/pkg/util/logger/logger_writer.go
@@ -15,7 +15,7 @@ import (
 
 const (
 	NumMessages   = 10 * 1024 // number of allowed log messages
-	STDOUT_FORMAT = "2006-01-02T15:04:05.000 "
+	STDOUT_FORMAT = "2006-01-02T15:04:05.000-07:00 "
 
 	LOG_INFO    = 2
 	LOG_DEBUG   = 1


steventine:
Would it be safer to make this change only when an explicit config flag is passed in on the command line to not risk breaking existing sinks?

i3149:
Closing this because its superseeded.

PR #861, opened 2025-12-31
fixing a bug with file tags 

i3149:
where the same value is different columns was missed. For example, 

```
1,C_DST_CUSTOMERNETWORK,222,FOO
2,C_SRC_CUSTOMERNETWORK,333,BAR
```

Will only show one value for the current setup. New system fixes. 

Added to test case. What I get for ignoring one of the values on input.


aschneider-kentik (approved):
LGTM

PR #899, opened 2026-05-19
Fixing a bug where nil threads in discovery will cause process to hang

### LINKED ISSUE #898: [Bug] Default discovery threads to 4 when unset to prevent silent deadlock
**Summary**

  - If threads is omitted from the discovery: section of snmp-base.yaml, conf.Disco.Threads is 0
  - make(chan bool, 0) creates a zero-capacity semaphore with no tokens, causing all doubleCheckHost goroutines to block indefinitely on <-ctl
  - wg.Wait() never returns — the container stays healthy but stops logging after Starting to check N ips from netbox with no error or panic

**Fix (disco.go)**
```golang
  if conf.Disco.Threads == 0 {
      conf.Disco.Threads = 4 
  }   
  ```

**How to reproduce**

Omit threads from the discovery: block in snmp-base.yaml and run with a Netbox config. The process will silently hang after fetching IPs from Netbox.

i3149:
Closes #898 

Fixes a case where a deadlock can happen because no one is actually processing the discovery.

### DIFF
--- a/pkg/inputs/snmp/disco.go
+++ b/pkg/inputs/snmp/disco.go
@@ -46,6 +46,11 @@ func Discover(ctx context.Context, log logger.ContextL, pollDuration time.Durati
 		return nil, err
 	}
 
+	if conf.Disco.Threads == 0 {
+		conf.Disco.Threads = 4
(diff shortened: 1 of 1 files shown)


alexraddas:
@i3149 a rockstar as always!

PR #901, opened 2026-05-26
Adding a vrf field for flow consumed by ktrans directly

i3149:


Mesverrum (dismissed):
Thanks, makes perfect sense to me

PR #904, opened 2026-06-11
Update New Relic API endpoints for JP region

jryburn:
The new endpoint URLs for Japan in newrelic need updating


i3149:
Closing because covered in #905

PR #912, opened 2026-06-25
logging export drops for otel now

### LINKED ISSUE #908: otel.no_block enhanced logging when chan is full
The otel.no_block option has been very useful for us to keep metrics flowing for some high cardinality locations 
https://github.com/kentik/ktranslate/issues?q=is%3Aissue%20otel

It would be great to drill in a little deeper to log what metricName's are getting dropped 
perhaps something like 

pkg/formats/otel/otel.go

```
   if f.config.NoBlockExport {
       select {
       case ch <- m:
       default:
           f.metrics.ExportDrops.Inc(1)
  +        f.Warnf("OTEL channel full, dropping sample for metric=%s", m.Name)
       }
   } else {
       ch <- m
   }
```

i3149:
Closes #908 

Better logging for otel.

### DIFF
--- a/pkg/formats/otel/otel.go
+++ b/pkg/formats/otel/otel.go
@@ -42,6 +42,7 @@ type OtelFormat struct {
 	trapLog      *OtelLogger
 	logTee       chan string
 	metrics      *OtelMetrics
+	warnFull     map[string]bool
 }
 
 const (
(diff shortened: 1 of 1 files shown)


i3149:
@dconnolly-sfdc, does this work for your use case? I'm nervous about logging at warn level in the hot loop of all exports.

dconnolly-sfdc:
@i3149 yes Ian, this will work for us

PR #919, opened 2026-07-16
Sanitize invalid UTF-8 in OTEL attributes to avoid batch drops

Mesverrum:
## Summary
- Add `kt.SanitizeUTF8` / `SanitizeUTF8WithHint`: leave valid UTF-8 alone; hex-encode invalid bytes (same idea as SNMP traps).
- Use it in the OTEL formatter for metric names and attribute keys/values so one bad SNMP string cannot fail the whole OTLP export.
- Point SNMP traps at the shared helper.

## Why
OTLP/gRPC marshaling rejects invalid UTF-8 and drops the entire metric batch. Commvault and similar devices often emit non-UTF-8 `ifAlias` / `sysName` / MIB strings.

## Test plan
- [x] `CGO_ENABLED=0 go test ./pkg/kt/ -run TestSanitizeUTF8 -v`
- [x] `CGO_ENABLED=0` compile of `pkg/formats/otel` and `pkg/inputs/snmp/traps`
- [ ] Run SNMP poller against a device with known bad UTF-8 aliases and confirm metrics still export (alias shows as hex)

### DIFF
--- a/pkg/formats/otel/otel.go
+++ b/pkg/formats/otel/otel.go
@@ -195,6 +195,10 @@ func (f *OtelFormat) To(msgs []*kt.JCHF, serBuf []byte) (*kt.Output, error) {
 		return nil, nil
 	}
 
+	for i := range res {
+		res[i].Name = kt.SanitizeUTF8(res[i].Name)
+	}
+
 	f.mux.RLock()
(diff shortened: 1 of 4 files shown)


i3149 (approved):
LGTM

PR #934, opened 2026-09-10
Synth error timeout mikek

mkrygeri:
This PR addresses the issue where "result type" of error or timeout are ignored. This adds a new metric to all other result types that allows users to track the state of a test if it is healthy, erroring or timing out. 
there is new documentation specifically for Firehose and Synthetics.

Another aspect to the Prometheus metrics was that the push metrics did not work for synthetics. This was fixed to reflect the same functionality of the scraper endpoint. The main difference id that this will push each result rather than a periodic scrape.


i3149 (approved):
LGTM!

PR #939, opened 2026-09-19
Allow custom per device rules

### LINKED ISSUE #937: File-based device lookup enrichment (inventory labels without rewriting user_tags)
## Summary

Add a **device-keyed lookup file** so SNMP metrics, traps, syslog, and flow can be enriched with inventory attributes (site, `edge_id`, `role`, `circuit_id`, …) without rewriting `user_tags` on every device in `snmp-base.yaml`.

This is **not** a MaxMind/`-geo` extension. Geo/ASN lookup is keyed on flow 5-tuple IPs. Device relationships are keyed on **`device_name` / `device_ip`** (the exporter / trap agent / syslog host). Using src/dst flow addresses would stamp parent/child labels on every conversation that merely transits an edge.

Related: #863 (NetBox → metric attributes) is the CMDB-specific version of the same idea. This issue is the generic file join so any exporter (NetBox, UA, a spreadsheet) can produce one document and ktranslate can apply it.

## Why `user_tags` is not enough

Today the supported path is:

```yaml
devices:
  leaf-br1:
    device_name: leaf-br1
    device_ip: 192.168.21.2
    user_tags:
      edge_id: spine1
      role: child
      circuit_id: WAN-HQ-BR1
      site: branch1
```

That works — `InitUserTags` / `SetMsgUserTags` already copy those onto SNMP, traps, and syslog when the device matches. It does **not** scale when:

- Inventory lives outside `snmp-base.yaml` (NetBox, UA, a circuit catalog).
- The same bag of labels should apply to **several containers** (SNMP poller, trap listener, syslog, flow) without duplicating tags in each generated devices file.
- You want to **reload the bag** without regenerating discovery output.

Workarounds today:

| Path | Gap |
|---|---|
| Stamp `user_tags` from a script at generate time | Brittle; catalog bounce to pick up changes |
| `-enricher` Starlark | No file I/O (`LoadFunc` only allows `json.star` / `math.star` / `time.star`); JCHF only — **syslog `logTee` never hits `doEnrichments`** |
| `-enricher` HTTP | Per-batch POST; not viable at high event volume |
| `-geo` / `-asn` | Wrong key (flow endpoint IPs, not device identity) |

Closest in-tree analog is `-application_map` (`pkg/util/rule`): load YAML → hashmap → stamp `CustomStr` in `doEnrichments`. We want that shape, keyed on **device identity**, and applied on **every signal** that already does a device match.

## Proposed interface

New flag, same style as `-application_map`:

```text
-device_lookup /path/to/device-lookup.yaml
```

Optional later: watch/reload the file (geo does not; `user_tags` already require a bounce). v1 can load at start.

### File format (v1)

YAML map. Each key is a **`device_name` or `device_ip`**. Values are string labels to copy onto matching telemetry.

```yaml
# device-lookup.yaml
#
# Join keys: device_name OR device_ip (exact match).
# Do not key on flow src/dst — those are clients, not the exporter.

spine1:
  site: hq
  edge_id: spine1
  role: parent
  circuit_id: ""          # parent may participate in several circuits; see notes

leaf-br1:
  site: branch1
  edge_id: spine1
  role: child
  circuit_id: WAN-HQ-BR1

leaf-br2:
  site: branch2
  edge_id: spine1
  role: child
  circuit_id: WAN-HQ-BR2

# Optional second index: same bag keyed by management / sampler IP
192.168.21.2:
  site: branch1
  edge_id: spine1
  role: child
  circuit_id: WAN-HQ-BR1
```

Lookup order on a record:

1. `device_name`
2. sampler / `device_ip` / trap agent address / syslog client IP
3. miss → leave telemetry unchanged (no error)

If both name and IP hit and disagree, **name wins**. Duplicate IPs across devices should log once at load.

### Where it must apply

This only pays off if it hits the same surfaces `user_tags` already do:

| Signal | Existing device match | Call site to hook |
|---|---|---|
| SNMP poll metrics | `devices[].device_ip` | After `InitUserTags` / when building metric attrs |
| SNMP traps | agent IP → device | Trap listener, same place tags are copied |
| Syslog | client IP / hostname → device (`SetMsgUserTags`) | `pkg/inputs/syslog` `formatMessage` — **do not rely on `doEnrichments`** |
| Flow / sFlow | sampler IP → catalog device | `handleInput` / `doEnrichments` via `DeviceName` |

Syslog is the easy miss: it is JSON on `logTee`, not JCHF, so a geo-style enricher in `handleInput` will never see it.

### What gets stamped

Copy lookup values into the same tag namespace as `user_tags` (`tags.<key>` on OTLP / NR, `CustomStr` on JCHF) so existing dashboards and Grafana inhibit/`equal=` labels keep working.

Example result on a trap from `leaf-br1`:

```json
{
  "device_name": "leaf-br1",
  "eventType": "KSnmpTrap",
  "tags.site": "branch1",
  "tags.edge_id": "spine1",
  "tags.role": "child",
  "tags.circuit_id": "WAN-HQ-BR1"
}
```

`user_tags` on the device remain authoritative for keys they already set (lookup fills **missing** keys only). That lets a device keep a one-off override without editing the shared file.

## Out of scope (v1)

- CIDR / prefix → site (could reuse Patricia later; EdgeID is a name hashmap).
- Keying on flow `src_addr` / `dst_addr`.
- HTTP fetch of the file (local path first; `git://` SNMP config already exists if we want that later).
- Starlark `load("device-lookup.yaml")` — a builtin would help scripts, but the flag should not require writing Starlark.
- Inferring parent/child from sysName regex. The file is the SoT; election belongs in the exporter that writes it.

## Implementation sketch

Follow `pkg/util/rule.NewRuleSet` + `application_map`:

1. Small helper, e.g. `pkg/util/devicelookup`, loads YAML into `map[string]map[string]string` (two indexes or one map that accepts both name and IP keys).
2. `Lookup(name, ip string) map[string]string`.
3. Call from:
   - device tag init (SNMP)
   - trap decorate
   - syslog `formatMessage` after device match
   - `doEnrichments` for flow JCHF (`msg.DeviceName`, sampler)
4. Tests: name hit, IP hit, name wins, miss, syslog JSON tags, trap JCHF `CustomStr`, device `user_tags` override.

Happy to follow with a PR once this shape looks right.

## Example operator workflow

NetBox / UA / a circuit catalog dumps `device-lookup.yaml`. ktranslate:

```text
ktranslate -snmp /etc/kt/snmp-base.yaml \
  -device_lookup /etc/kt/device-lookup.yaml \
  ...
```

Grafana alert inhibit then uses labels that already exist on the series/log:

- parent: `role=parent`
- child: `role=child`
- equal: `edge_id` or `circuit_id`

No PromQL join and no per-device `user_tags` rewrite.

i3149:
Closes #937 

Adds in an optional flag `-user_device_rule_path`. This Loads a yaml file as defined in #937 and sets tags accordingly. Allows local files, http(s):// and s3:// style values for the yaml file.


Mesverrum:
Thanks for picking this up — this is the shape I was hoping for: a sidecar YAML keyed on `device_name` / `device_ip`, merged into `InitUserTags` so SNMP / traps / syslog / flow keep using the existing tag copy path, and `user_tags` on the device still win.

A couple of things before merge, then a small follow-up if you're up for it.

### Same PR (would block a default install)

`InitUserDeviceRuleSet` is always called from `NewKTranslate`, and `LoadFile("")` falls through to `os.ReadFile("")`. Unless I'm missing a guard, a process that never sets `-user_device_rule_path` will fail to start. An empty-path early return would match how `-application_map` already works.

Related: a YAML key that looks like a CIDR (`192.168.21.0/24`) currently fails `net.ParseIP` and lands in `deviceNameUserTags`, so it silently never matches an IP. Fine to leave CIDR out of this PR, but worth not treating `/` keys as names (skip + log, or park them for later).

Minor: `Device.InitUserTags(..., nil, nil)` on the missing-device path will panic if `allUserTags` is nil and `serviceName != "ktranslate"`.

### Follow-up (not asking you to do this here)

You asked on the issue about IP ranges. I'd keep exact name → exact IP, then **longest-prefix CIDR** as a third index, same property bags:

```yaml
192.168.21.0/24:
  site: branch1
  edge_id: spine1
```

Lookup stays load-time (`GetUserDeviceRuleSet` at init, not per packet), so a sorted `[]net.IPNet` is enough; `IPAddressRules` / patricia is there if you prefer LPM for free. Happy to send that as a follow-up PR after this lands, or push a commit onto this branch if you'd rather keep it together — your call.

Mesverrum:
Follow-up from the earlier review: I ran this branch locally (`e675953`) and added table-driven-ish edge-case tests around empty / odd YAML.

### What fails on the PR as-is

| Case | Result |
|---|---|
| `InitUserDeviceRuleSet("")` | `open : no such file or directory` — default install with the flag unset cannot start |
| YAML key `192.168.21.0/24` | Treated as a **device name**; lookup by that string hits, lookup by an IP in the subnet does not |
| YAML key `""` | Stored and returned for empty `device_name` |
| `"  leaf-br1  "` | No match on `leaf-br1` (no trim) |
| `Device.InitUserTags("ktranslate-flow", nil, nil)` | **panic**: assignment to entry in nil map |
| `SnmpDeviceConfig` with `UserTags: nil` + `serviceName == "ktranslate"` + defaults | nil-map write when merging the lookup bag |

Happy path still looks good: name wins over IP, IP match on name miss, empty file / `{}` loads as zero rules, invalid YAML errors, IPv6 exact match, device `user_tags` still override the file.

### Odd data (current decode — I would keep this)

- `circuit_id: ""` and YAML `null` both become `""` and **are stamped**. That matches your parent-circuit example.
- Unquoted `vlan: 100` / `core: true` coerce to `"100"` / `"true"` (yaml.v3). Fine.
- Nested maps (`site: { name: branch1 }`) fail unmarshal — good.
- Empty file and `{}` are a no-op.

I would **not** drop empty-string values; people will use `circuit_id: ""` on purpose.

### Suggested guards (not pushed to your branch)

Empty path no-op, skip blank keys, skip CIDR keys with a warn (until LPM exists), normalize IP keys via `net.ParseIP().String()`, allocate empty maps before write. Patch against this branch:

```diff
diff --git a/pkg/kt/device_types.go b/pkg/kt/device_types.go
@@ -126,6 +126,9 @@ type Column struct {
 }
 
 func (d *Device) InitUserTags(serviceName string, tags map[string]string, defaults map[string]string) {
+	if tags == nil {
+		tags = map[string]string{}
+	}
 	d.allUserTags = tags
```

```diff
diff --git a/pkg/kt/snmp.go b/pkg/kt/snmp.go
@@ -811,10 +811,10 @@ func (d *SnmpDeviceConfig) InitUserTags(serviceName string, defaults map[string]string) {
 		panic(...)
 	}
 
+	if d.UserTags == nil {
+		d.UserTags = map[string]string{}
+	}
 	if serviceName != "ktranslate" {
-		if d.UserTags == nil {
-			d.UserTags = map[string]string{}
-		}
 		d.UserTags["container_service"] = serviceName
 	}
```

```diff
diff --git a/pkg/util/rule/rule.go b/pkg/util/rule/rule.go
@@ func InitUserDeviceRuleSet
+	if strings.TrimSpace(rulePath) == "" {
+		return nil
+	}
@@ classify keys
+		tk = strings.TrimSpace(tk)
+		if tk == "" { continue }
+		if ip := net.ParseIP(tk); ip != nil {
+			deviceIpUserTags[ip.String()] = mm
+			continue
+		}
+		if _, _, err := net.ParseCIDR(tk); err == nil {
+			log.Warnf("Skipping CIDR user device rule %s (exact name/IP only in this change)", tk)
+			continue
+		}
+		deviceNameUserTags[tk] = mm
@@ GetUserDeviceRuleSet
+	trim name; normalize IP with net.ParseIP().String() before map lookup
```

With those three files patched, `go test ./pkg/util/rule/ ./pkg/kt/ ./pkg/inputs/snmp/metadata/` is green, including your existing `TestToFlowsWithRuleSet`.

Happy to send this as a tiny follow-up PR after you merge, or drop a commit on this branch if you want it in #939. CIDR longest-prefix still feels like a separate change.

Mesverrum (changes requested):
Thanks for the follow-up — I re-tested `e68ee79` locally (`go test ./pkg/util/rule/ ./pkg/kt/ ./pkg/inputs/snmp/metadata/`).

### Fixed since the last pass
- Empty flag no longer breaks default boot (`NewKTranslate` skips `InitUserDeviceRuleSet` when the path is empty).
- Nil `tags` on `Device.InitUserTags` no longer panics; SNMP `UserTags == nil` is allocated before merging defaults.
- Flag help text is cleaned up.
- **CIDR works** for a single prefix: `192.168.21.0/24` matches `192.168.21.16` and is no longer stored as a device name. `TestToFlowsWithRuleSet` covers name / IP / CIDR.
- Device `user_tags` still win; empty/`{}` YAML is fine; ints/bools coerce to strings.

### Remaining issue (would request this before merge)

**Overlapping CIDRs do not use longest-prefix match.** `deviceCidrUserTags` is a `map[*net.IPNet]…` and lookup returns the **first** `Contains` hit, which is map iteration order.

I ran `TestCIDRLongestPrefixWins` 50 times with:

```yaml
192.168.0.0/16:   { site: wide }
192.168.21.0/24:  { site: branch1 }
```

Most runs stamped `site=wide` for `192.168.21.16`. A single-prefix catalog is fine; two nested site prefixes are not deterministic.

Fix is small: keep a slice, sort by mask length descending, or reuse `IPAddressRules` / patricia `FindDeepestTag` (payload as an index into the bags). Please add that case next to `TestToFlowsWithRuleSet`.

### Nice-to-have (not blocking if CIDR catalogs stay disjoint)

- `InitUserDeviceRuleSet("")` still `LoadFile("")`s — callers guard it; an early return inside Init would be safer.
- YAML key `""` still matches empty `device_name`.
- `"  leaf-br1  "` does not match `leaf-br1` (no trim).
- IPv6 exact keys are not normalized: `2001:db8::1` misses `2001:0db8::0001`.
- Device tag tests only assert `len(map)`, not values.

Happy to approve once overlapping CIDRs pick the longest prefix (or you document “first match, keep prefixes disjoint” and skip the flake).

i3149:
Give that a shot. I hit your list and remembered how the patricia library works.

Mesverrum (changes requested):
Nice — patricia `FindDeepestTag` was the right call. I pulled `c588b1d` and re-ran locally.

### Confirmed
- Empty path in `InitUserDeviceRuleSet` is a no-op.
- Keys are trimmed; blank YAML keys skipped; IPv6 exact keys normalized (`2001:db8::1` matches `2001:0db8::0001`).
- Overlapping CIDRs: `TestCIDRLongestPrefixWins` (`/16` vs `/24` vs `/8`) passed **50/50**. Your expanded `TestToFlowsWithRuleSet` (`middle` / `small` / `large`) also passes.
- Name still wins over IP; device `user_tags` still win.

### One leftover that would hit default installs

`GetUserDeviceRuleSet` always calls `deviceRules.Check(ip)` when the device has an IP. If `-user_device_rule_path` is unset, `InitUserDeviceRuleSet` never runs and `deviceRules` stays `nil` → **nil pointer** in `StringAddressRule.Check`.

`parseConfig` always does `GetUserDeviceRuleSet(device.DeviceName, device.DeviceIP)`, so SNMP with a real `device_ip` and no sidecar file would panic on start. Existing tests miss it because they either Init first or leave `DeviceIP` empty (`ParseIP("")` is nil, so Check is skipped).

```go
if deviceRules != nil {
    match := deviceRules.Check(nip)
    ...
}
```

or initialize `deviceRules = NewStringAddressRule()` in the package `var` block / empty-path Init.

Happy to approve as soon as that nil guard is in.

i3149:
Added a test and guard to cover the nil case.

Mesverrum (approved):

